PR #1, opened 2025-07-01
Initial ADR

reinkrul:


stevenvegt (changes requested):
why in the template dir!!??!

Issue #19, opened 2025-07-22
Setup github actions for building en testing


stevenvegt:
Duplicate the logic from the nuts-node repo as much as possible.

PR #25, opened 2025-07-22
#19: basic application which can be deployed

reinkrul:
I also quickly dropped in zerolog (https://github.com/rs/zerolog) so we can log in JSON format.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,33 @@
+FROM golang:1.24.4-alpine AS builder
+
+ARG TARGETARCH
+ARG TARGETOS
+
+ENV GOPATH /
+
+COPY go.mod .
+COPY go.sum .
+RUN go mod download && go mod verify
+COPY . .
+
+RUN mkdir /app
+RUN GOOS=$TARGETOS GOARCH=$TARGETARCH go build -o /app/bin .
+
+# alpine
+FROM alpine:3.22.0
+RUN apk update \
+  && apk add --no-cache \
+             tzdata \
+             curl
+COPY --from=builder /app/bin /app/bin
+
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+COPY go.mod .
+COPY go.sum .
+RUN go mod download && go mod verify
+COPY . .
+
+RUN mkdir /app
+RUN GOOS=$TARGETOS GOARCH=$TARGETARCH go build -o /app/bin .
+
+# alpine
+FROM alpine:3.22.0
```
dmeijboom, line 17: Nitpick: I prefer distroless (e.g. FROM gcr.io/distroless/static-debian12), it's smaller (~ 2MB) and already configures a nonroot user and doesn't have performance issues in the memory allocator (musl libc it's allocator can be slow). It doesn't have curl so it would require removing the HEALTHCHECK.

dmeijboom (approved):
LGTM

Dockerfile
reinkrul: Can we install curl on it, like we do here? It's really nice to have a health check when running in docker (especially for local development), and I sometimes use it to debug something in a running cluster (although that's probably a hack..)

dmeijboom: I guess this is a decent solution: https://github.com/GoogleContainerTools/distroless/issues/183#issuecomment-571723446 (copies `wget` from busybox)

confiks: I think using `alpine` for now is fine. Copying wget from busybox is also an option, but I wouldn't want to do it through `ADD https://`

confiks (approved):

PR #49, opened 2025-08-11
Move HAPI FHIR server to port 8081

Dirklectisch:
Tiny PR for resolving a conflict between the FHIR server running in the docker-compose container and the knooppunt app. They were both using the same port and thus couldn't be started at the same time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,4 @@
 .idea
 
 # Binaries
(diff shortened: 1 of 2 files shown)


Dirklectisch:
Changed to 7050

reinkrul (approved):

PR #52, opened 2025-08-12
feat: Add GitHub Action to generate C4 diagrams

### LINKED ISSUE #32: Automagically create c4 diagrams in github action
In the /docs directory there is a C4 diagram.

Using a github action with the structurizr docker container to create a set of png's.
These png's should be referenced in a readme

See docs: https://docs.structurizr.com/lite/workflow

JorisHeadease:
Adds automated generation of C4 architecture diagrams from Structurizr DSL files

## Changes
- Add GitHub Action workflow that generates PNG diagrams using structurizr-to-png
- Update README to display the generated diagrams
- Workflow runs on changes to the DSL file and commits the generated images

Closes #32


reinkrul (approved):
Nice! Just checking, if the same PNGs are generated twice, are they binary equal (for the `git diff`)?

reinkrul:
Note that this might also be a pre-commit-hook instead, then you also have the changed images as part of the PR

JorisHeadease:
> Nice! Just checking, if the same PNGs are generated twice, are they binary equal (for the `git diff`)?

Ah, good catch! This probably won't work due to png metadata. I've completely removed it, as the action already only triggers on changes inside the diagram file or the GitHub action. So it's redundant and probably incorrect

JorisHeadease:
Closing PR as mentioned at [this](https://github.com/nuts-foundation/nuts-knooppunt/issues/32#issuecomment-3223031024) comment.

> Ticket closed because the group concludes it is a bit over-engineering because we only have around 5 C4-diagrams.

reinkrul:
Let's keep the branch, I suspect we might still want it soon...

PR #53, opened 2025-08-12
#23: Embed Nuts node

### LINKED ISSUE #23: Embed Nuts node in application
We embed applications in the Knooppunt app as subsystems. Do this for the Nuts node. Each "subsystem" will be mounted as a route group while public/private endpoints are separated.

reinkrul:
This adds a Nuts node component, which binds its own HTTP interfaces (available from localhost only, on random ports). The component then registers a reverse proxy on `/nuts` to forward HTTP requests to the Nuts node. It can be configured using `config.nuts.yaml` or environment variables.

It can be disabled with an environment variable (to be replaced with actual config, e.g. `koanf`, at some point).

Closes #23

TODO:
- [x] Route well-known endpoints: `/.well-known/nuts/<some-path>` has to be handled by Nuts as `/.well-known/nuts/<some-path>`


component/nutsnode/proxy.go
JorisHeadease, line 1: We should add some unit tests that prove the rewriting keeps working

lib/must/util.go
```diff
@@ -0,0 +1,11 @@
+package must
```
confiks, line 1: What's the purpose of the `must` package? That it panics? Why not just use `url.Parse` as-is and handle any errors?

component/nutsnode/util.go
```diff
@@ -0,0 +1,49 @@
...
+	return logrus.AllLevels
+}
+
+func (a logrusZerologBridgeHook) Fire(entry *logrus.Entry) error {
+	entry.Data["component"] = "nutsnode"
+	fields := map[string]interface{}(entry.Data)
+	logger := zerolog.DefaultContextLogger
+	switch entry.Level {
+	case logrus.DebugLevel:
+		logger.Debug().Fields(fields).Msg(entry.Message)
```
JorisHeadease, line 25:
```suggestion
	switch entry.Level {
	case logrus.TraceLevel:
		logger.Trace().Fields(fields).Msg(entry.Message)
	case logrus.DebugLevel:
		logger.Debug().Fields(fields).Msg(entry.Message)
```

config/init/main.go
```diff
@@ -0,0 +1,59 @@
...
+	"os"
+	"strings"
+)
+
+func main() {
+	if len(os.Args) < 2 {
+		panic("Usage: " + os.Args[0] + " <internal API base path>")
+	}
+	internalAPI := os.Args[1]
+	tenantName := "orgA"
```
JorisHeadease, line 17: Why is this hard coded?

config/init/main.go
JorisHeadease, line 1: I don't really understand why this file is in a generic place like `config/init`. The file only seems to add new nuts subjects from the CLI; it doesn't really seem related to configuration?

lib/netutil/util.go
JorisHeadease, line 1: Maybe rename this file/folder to explicitly inform devs that it's a "testing" util? This function should probably be only used in unit tests. And as well, perhaps we could even use build tags and exclude the code from the built binary?

JorisHeadease (changes requested):

Issue #85, opened 2025-09-01
mCSD Update Client: generate new resource ID for copy in organization Query Directory


reinkrul:
Currently, mCSD resources keep the same resource ID as from the root/discovery Admin Directory, but in local query dirs, they should have their own IDs

PR #86, opened 2025-09-01
#85: Generate new IDs when updating mCSD to local query directory

reinkrul:


component/mcsd/updater.go
```diff
@@ -48,31 +50,80 @@ func buildUpdateTransaction(ctx context.Context, tx *fhir.Bundle, entries []fhir
...
+		if localResourceID == "" {
+			localResourceID = generateLocalID()
+			remoteRefToLocalRefMap[remoteLocalRef] = localResourceID
+		}
+		resource["id"] = localResourceID
+		if err := normalizeReferences(resource, remoteRefToLocalRefMap); err != nil {
+			return nil, fmt.Errorf("failed to normalize references in entry #%d: %w", i, err)
+		}
+
 		resourceJSON, _ := json.Marshal(resource)
```
JorisHeadease, line 68: We probably do not want to swallow marshalling errors?

component/mcsd/updater.go
```diff
@@ -48,31 +50,80 @@ func buildUpdateTransaction(ctx context.Context, tx *fhir.Bundle, entries []fhir
...
+		// TODO: We should scope resource IDs to the source (e.g. by prefixing with the source URL or a hash thereof), because when syncing from multiple sources, IDs may collide.
+		remoteLocalRef := resourceType + "/" + resource["id"].(string)
+		localResourceID := remoteRefToLocalRefMap[remoteLocalRef]
+		if localResourceID == "" {
+			localResourceID = generateLocalID()
+			remoteRefToLocalRefMap[remoteLocalRef] = localResourceID
+		}
+		resource["id"] = localResourceID
+		if err := normalizeReferences(resource, remoteRefToLocalRefMap); err != nil {
+			return nil, fmt.Errorf("failed to normalize references in entry #%d: %w", i, err)
```
JorisHeadease, line 65: Maybe good to add the resource ref in the error to ease troubleshooting

component/mcsd/updater.go
```diff
@@ -48,31 +50,80 @@ func buildUpdateTransaction(ctx context.Context, tx *fhir.Bundle, entries []fhir
...
-		//}
-		//resource["id"] = localResourceID
+		// TODO: If the resource already exists, we should look up the existing resource's ID
+		// TODO: We should scope resource IDs to the source (e.g. by prefixing with the source URL or a hash thereof), because when syncing from multiple sources, IDs may collide.
+		remoteLocalRef := resourceType + "/" + resource["id"].(string)
+		localResourceID := remoteRefToLocalRefMap[remoteLocalRef]
+		if localResourceID == "" {
+			localResourceID = generateLocalID()
+			remoteRefToLocalRefMap[remoteLocalRef] = localResourceID
+		}
```
JorisHeadease, line 62: The `localResourceID` contains `urn:uuid:<UUID>`, which is a "bundle reference". When this bundle is POST'ed, the server will replace that bundle ref with the "actual" local id. So we'd need to set this reference from the response bundle, instead of this value I'd think?

JorisHeadease (changes requested):

component/mcsd/updater.go
reinkrul: we unmarshalled it, so it can be marshalled again, but returning the err won't hurt here

component/mcsd/updater.go
JorisHeadease: Ahhh you're absolutely right! I assumed it was a "global" in-memory map. Never mind my comment then!

PR #99, opened 2025-09-02
Apply Bootstrap 4 to mCSD Admin UI

reinkrul:


component/mcsdadmin/templates/organization_list.html
```diff
@@ -1,27 +1,34 @@
...
+                <th scope="col">Active</th>
+            </tr>
+            </thead>
+            <tbody>
+            {{range .Organizations}}
+            <tr>
+                <th scope="row">{{ .Name }}</th>
+                <td>{{.Type}}</td>
+                <td>
+                    {{if .Active}}
```
Dirklectisch, line 22: This is a bug I think, .Active is a string it's zero value is empty string so the status here will always show the active badge. I'll implement a fix.

Dirklectisch (approved):
I fixed a bug with the display of one of the fields. Other than that I would have preferred to just have pure CSS styling. I would think Copilot would have been able to write that as well. Now we have increased the tower op abstraction by having AI use Bootstrap.

In any case it works and looks more presentable so I'll merge it in. 👍

Issue #90, opened 2025-09-02
Write Nuts KNPT deployment manual


jorritspee:
Things to describe:

- Docker image documentation
- Configuration documentation
      - Nuts-node is disabled for now
      - Configure administration directory connection
      - Configure query directory connection
      - Configure root directory connection
- Describe how to make your services discoverable

jorritspee:
Okay, I don't know how this is done normally. It looked a bit messy to me but I am not the target audience. Let's just not change anything until a developer has a strong opinion about it ;-).

PR #105, opened 2025-09-04
#90 deployment guide

Dirklectisch:
I wrote a short deployment guide in which we will document what implementors need to do to start using Nuts Knooppunt.

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -3,7 +3,7 @@
 Nuts Knooppunt is provided as a single docker image that should be deployed within a participating organisation's
 network. We will update this guide as we implement the different generic functions.
 
-Knooppunt is available from the Github Container Registry.
+Knooppunt is available from the GitHub Container Registry.
 
 https://github.com/nuts-foundation/nuts-knooppunt/pkgs/container/nuts-knooppunt
 
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -21,12 +21,11 @@ providing data according to the mCSD profile.
 A multi tenant HAPI server can be used for hosting both the admin and query directory. We recommend to keep this data
 separate, but you can choose to combine the data in a single tenant if so desired.
 
-The root directory will be the LRZA directory provided by the ministry of health (VWS). During testing Nuts will provide
-an example root directory which is available on this URL:
+The root directory will be the LRZA directory provided by the ministry of health (VWS). During preliminary testing, an example root directory is available on this URL:
 
 https://knooppunt-test.nuts-services.nl/lrza/mcsd
 
-Please notify Nuts if you would like to make a directory discoverable through our example
+Please get in contact if you would like to make your Administration Directory discoverable through our example
 root directory.
 
 For testing purposes your admin directory should be reachable through the public internet. The production scenario aims
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -1,0 +1,28 @@
+# Deployment
+
+Nuts Knooppunt is provided as a single docker image that should be deployed within a participating organisations
+network. We will update this guide as we implement the different generic functions.
+
+Knooppunt is available from the Github Container Registry.
+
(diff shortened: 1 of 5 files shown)


DEPLOYMENT.md
```diff
@@ -0,0 +1,35 @@
+# Deployment
+
+Nuts Knooppunt is provided as a single docker image that should be deployed within a participating organisations
+network. We will update this guide as we implement the different generic functions.
+
+Knooppunt is available from the Github Container Registry.
+
+https://github.com/nuts-foundation/nuts-knooppunt/pkgs/container/nuts-knooppunt
+
+Knooppunt embeds a Nuts Node which is disabled by default.
```
reinkrul, line 10:
```suggestion
Knooppunt embeds a Nuts Node which is enabled by default.
```

DEPLOYMENT.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Addressing
+
+To participate in the addressing generic function Knooppunt will need to connect to three different FHIR services
+providing data according to the mCSD profile.
+
+1. A root directory keeping authoritative references to a set of admin directories
+2. An admin directory used to publish data such as endpoints
+3. A query directory where the synchronisation process will place data received from other organisations
```
reinkrul, line 19:
```suggestion
To participate in the addressing generic function Knooppunt will connect to several FHIR APIs to synchronize data, according to the mCSD profile:
- a Root Administration Directory, authoratitive on care organizations and their Administration Directory endpoints,
- Administration Directories of care organizations, discovered through the Root Administration Directory.
- a local Query Directory where the synchronisation process will put data received from other organisations

For your own Knooppunt, you need to:
- provide a FHIR API for the Query Directory
- provide a FHIR API for the Administration Directory, containing your care organization registrations, that other Knooppunt instances can query.
  - You can use the embedded mCSD Admin Editor web application (`/mcsdadmin`) to maintain this directory.
- configure the Root Administration Directory
```

DEPLOYMENT.md
```diff
@@ -0,0 +1,35 @@
...
+The root directory will be the LRZA directory provided by the ministry of health (VWS). During testing Nuts will provide
+an example root directory which is available on this URL:
+
+https://knooppunt-test.nuts-services.nl/lrza/mcsd
+
+Please notify Nuts if you would like to make a directory discoverable through our example
+root directory.
+
+For testing purposes your admin directory should be reachable through the public internet. The production scenario aims
+to utilise mTLS for trusted communication.
```
reinkrul, line 33: I wouldn't say "notify Nuts";

```suggestion
The root directory will be the LRZA directory provided by the ministry of health (VWS). During preliminary testing, an example root directory is available on this URL:

https://knooppunt-test.nuts-services.nl/lrza/mcsd

Please get in contact if you would like to make your Administration Directory discoverable through our example
root directory.

For testing purposes your admin directory should be reachable through the public internet. The production scenario aims
to utilise mTLS for trusted communication.
```

reinkrul (changes requested):

DEPLOYMENT.md
Dirklectisch: We wrote in #90 that it should be disabled by default. But I do see we have configured it to be enabled now. Shouldn't we disable it by default until the implementor decides to use it?

reinkrul: Sure, but we need to change that.

Dirklectisch: Done!

config/knooppunt.yml
```diff
@@ -22,4 +22,4 @@ mcsdadmin:
 # Nuts node configuration  
 nuts:
   # Whether to enable the Nuts node component
-  enabled: true
+  enabled: false
```
reinkrul, line 25: I would just change the default in `DefaultConfig()`, and keep this on `true`, for easier testing

Dirklectisch: Makes sense 🫣

PR #118, opened 2025-09-09
Fix Docker healthcheck

reinkrul:


gasperr (approved):
I've encountered the same issue trying to set these things up so I can confirm the change fixes this issue with healthcheck.

Issue #121, opened 2025-09-09
mCSD Update Client: use conditional update _source to appy changes


reinkrul:
This way, we support local resource IDs without having to do lookups

PR #122, opened 2025-09-09
#121: Conditional update with _source

reinkrul:


JorisHeadease (commented):
I didn't finish the PR fully as we probably first need to decide on the approach. To keep it as is, or simplify the code with [conditional references](https://hl7.org/fhir/http.html#vs-refs). My main question is if it replaces both the chain handlers (the bundle references) and the local FHIR search resolution (second chain handler). Pretty sure if does the latter, not sure in the prior. According to Gasper it works well in Firely and HAPI, but we're not sure on others like Azure

PR #123, opened 2025-09-10
feat(mcsd): add deduplication for _history responses

JorisHeadease:
Fix HAPI-0534 error by deduplicating resources from _history queries. Keep only most recent version based on versionId with lastUpdated fallback. Prevents duplicate resource IDs in transaction bundles.

Add comprehensive tests and TODO comments for future _source support.


reinkrul (changes requested):
small things

PR #125, opened 2025-09-10
Check for payload type instead of connection type during sync process

Dirklectisch:
When determining if a resource is a mcsd admin directory we should be using the payload type instead of the connection type. This PR adds the new values and updates the tests accordingly.


component/mcsd/updater.go
```diff
@@ -57,8 +57,13 @@ func buildUpdateTransaction(tx *fhir.Bundle, entry fhir.BundleEntry, allowedReso
 			if err := json.Unmarshal(entry.Resource, &endpoint); err != nil {
 				return "", fmt.Errorf("failed to unmarshal Endpoint resource: %w", err)
 			}
+
 			// Import mCSD directory endpoints even from discoverable directories
-			doSync = coding.EqualsCode(endpoint.ConnectionType, coding.MCSDConnectionTypeSystem, coding.MCSDConnectionTypeDirectoryCode)
+			var payloadCoding = fhir.Coding{
```
reinkrul, line 62: could you move this as a constant (well it can't be a constant, but declared once, at least) to `coding` package?

Or you could make a defacto constant by making a function for it that returns this

component/mcsdadmin/valuesets/endpoint-payload-type.json
```diff
@@ -286,5 +286,13 @@
...
     "code": "urn:hl7-org:sdwg:ccda-nonXMLBody:1.1",
     "display": "For documents following C-CDA constraints using a non structured body. "
+  },
+  {
+    "code": "http://nuts-foundation.github.io/nl-generic-functions-ig/CapabilityStatement/nl-gf-admin-directory-update-client",
+    "display": "Care Services Directory for Update Client"
+  },
+  {
+    "code": "http://nictiz.nl/fhir/CapabilityStatement/eOverdracht-servercapabilities",
+    "display": "Transfer of Care - eOverdracht Server"
```
reinkrul, line 296: can we order these, so that these 2 end up on top in mCSD Admin? Just for the UX

lib/coding/constants.go
```diff
@@ -2,4 +2,5 @@ package coding
 
 const URANamingSystem = "http://fhir.nl/fhir/NamingSystem/ura"
 const MCSDConnectionTypeSystem = "http://fhir.nl/fhir/NamingSystem/endpoint-connection-type"
```
reinkrul, line 4: I think this can go as well?

lib/coding/util.go
```diff
@@ -8,3 +8,21 @@ func EqualsCode(coding fhir.Coding, system string, value string) bool {
 	return coding.System != nil && *coding.System == system &&
 		coding.Code != nil && *coding.Code == value
 }
+
+func CodableIncludesCode(codable fhir.CodeableConcept, code fhir.Coding) bool {
+	for _, c := range codable.Coding {
+		if *c.System == *code.System && *c.Code == *code.Code {
```
reinkrul, line 14: they can be `nil`, which would lead to panic due to nil dereference

lib/coding/util.go
```diff
@@ -8,3 +8,21 @@ func EqualsCode(coding fhir.Coding, system string, value string) bool {
 	return coding.System != nil && *coding.System == system &&
 		coding.Code != nil && *coding.Code == value
 }
+
+func CodableIncludesCode(codable fhir.CodeableConcept, code fhir.Coding) bool {
```
reinkrul, line 12: can you make a unit test for this?

test/testdata/go.mod
```diff
@@ -4,6 +4,7 @@ go 1.24.4
 
 require (
 	github.com/SanteonNL/go-fhir-client v0.5.0
+	github.com/nuts-foundation/nuts-knooppunt v0.1.1
```
reinkrul, line 7: we don't want this, because;
- it's a circular dependency now
- it makes testing slow, since the knooppunt-tests use this testdata package, which then goes on to download some (outdated) version of nuts-knooppunt

If we like the util funcs you added, you can add them to the `zorgbijjou/golang-fhir-models` library in the `caramel` package (because if you put syntactic sugar on FHIR/FIRE, you get caramel, you know...)

PR #151, opened 2025-09-16
#135: fix mCSD Update Client meta.source

### LINKED ISSUE #135: mCSD Update Client: meta.source is incorrect when syncing from remote LRZa
It works in local tests, but when syncing from remote LRZa:

```
      "meta": {
        "versionId": "1",
        "lastUpdated": "2025-09-12T09:51:17.564+00:00",
        "source": "Endpoint/Endpoint/4"
      }
```

reinkrul:
Fixes #135

### DIFF
--- a/component/mcsd/test/org1_dir_history_response-page2.json
+++ b/component/mcsd/test/org1_dir_history_response-page2.json
@@ -6,7 +6,7 @@
   "type": "history",
   "entry": [
     {
-      "fullUrl": "{{ORG1_DIR_BASEURL}}/Endpoint/ep-2",
+      "fullUrl": "Endpoint/ep-2",
       "resource": {
         "resourceType": "Endpoint",
         "id": "ep-2",
--- a/component/mcsd/updater.go
+++ b/component/mcsd/updater.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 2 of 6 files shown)


component/mcsd/test/org1_dir_history_response-page2.json
```diff
@@ -6,7 +6,7 @@
   "type": "history",
   "entry": [
     {
-      "fullUrl": "{{ORG1_DIR_BASEURL}}/Endpoint/ep-2",
+      "fullUrl": "Endpoint/ep-2",
```
JorisHeadease, line 9: Maybe we can change the page 2 fullUrls to version-specific URLs (relative and absolute) so we ensure that also keeps working

component/mcsd/updater.go
```diff
@@ -78,12 +79,16 @@ func buildUpdateTransaction(tx *fhir.Bundle, entry fhir.BundleEntry, allowedReso
 		return resourceType, nil
 	}
 
-	updateResourceMeta(resource, *entry.FullUrl)
+	originalResourceID, ok := resource["id"].(string)
```
JorisHeadease, line 82: We can use the fhir util with `ExtractResourceInfo` for the id and resourceType etc, might need to extend the logic for the `updateResourceMeta()` though

PR #166, opened 2025-09-24
feat(bsn): add proof-of-concept transport token utils

JorisHeadease:
This PR just includes the basic functions used by the "Knooppunten". It also contains the functions for the NVI. The latter should be removed, but we mainly wanted to prove the principle in a single test.

  **Knooppunt functions:**
  - `CreateTransportToken(bsn int, audience string)` - BSN → unique transport token
  - `BSNFromTransportToken(token string)` - Extract BSN from token

  **NVI functions (to be removed later):**
  - `TransportTokenToPseudonym(token string)` - Token → consistent pseudonym
  - `PseudonymToTransportToken(pseudonym string, audience string)` - Pseudonym → new org-specific token

  **Testing:**
  - Full test suite validating the complete workflow
  - Some edge cases

  **Note:** Uses simple XOR encryption (proof-of-concept only)

### DIFF
--- a/lib/bsnutil/util.go
+++ b/lib/bsnutil/util.go
@@ -1,0 +1,205 @@
+package bsnutil
+
+import (
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/binary"
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+const (
+	tokenPrefix        = "token-"
+	pseudonymPrefix    = "ps-"
+	minTokenLength     = len(tokenPrefix) + 1     // "token-" + at least 1 char
+	minPseudonymLength = len(pseudonymPrefix) + 1 // "ps-" + at least 1 char
+
+	// Input validation limits
+	maxAudienceLength = 255
+	minBSN            = 100000000
+	maxBSN            = 999999999
+)
+
+// CreateTransportToken creates a transport token from BSN and audience using simple XOR transformation.
+// Each transport token is unique (includes random nonce) but contains the same encrypted BSN.
+// This ensures transport tokens cannot be tracked while the NVI can always generate the same pseudonym.
+//
+// Parameters:
+//   - bsn: The Dutch social security number (BSN)
+//   - audience: The identifier for the organization/audience receiving the token
+//
+// Returns a transport token in format: "token-{audience}-{transformedBSN}-{nonce}"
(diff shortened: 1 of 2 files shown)


lib/bsnutil/util.go
```diff
@@ -0,0 +1,202 @@
...
+// CreateTransportToken creates a transport token from BSN and audience using simple XOR transformation.
+// Each transport token is unique (includes random nonce) but contains the same encrypted BSN.
+// This ensures transport tokens cannot be tracked while the NVI can always generate the same pseudonym.
+//
+// Parameters:
+//   - bsn: The Dutch social security number (BSN)
+//   - audience: The identifier for the organization/audience receiving the token
+//
+// Returns a transport token in format: "token-{audience}-{transformedBSN}-{nonce}"
+func CreateTransportToken(bsn int, audience string) (string, error) {
```
reinkrul, line 33: I'd make BSN a `string`, as it fixed length, at some systems may use prefix-zeroes

lib/bsnutil/util.go
```diff
@@ -0,0 +1,202 @@
...
+	key := generateSimpleKey(pseudonymHolder)
+	originalBSN := simpleXOR(transformedBSN, key)
+
+	// Create new token with the target audience
+	return CreateTransportToken(originalBSN, audience)
+}
+
+// validateBSN checks if the BSN is within valid range for Dutch social security numbers.
+func validateBSN(bsn int) error {
+	if bsn < minBSN || bsn > maxBSN {
```
reinkrul, line 160: I don't think we need to validate this here, since we can work with a simple plaintext, which can be whatever data (no need to check business rules here)

lib/bsnutil/util_test.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+func TestCompleteWorkflow(t *testing.T) {
+	// Test the complete federated health workflow from the architecture diagram
+	bsn := 123456789
+
+	// LOCALIZATION: Knooppunt A registers DocumentReference
+	// Step 1: Knooppunt A creates transport token with audience="nvi"
+	tokenFromA, err := CreateTransportToken(bsn, "nvi")
+	if err != nil {
```
reinkrul, line 189: can you use `require.NoError()` for shorter tests?

reinkrul (approved):

PR #169, opened 2025-09-25
ADR: Solution Architecture

reinkrul:


reinkrul:
> But this does not necessarily also need to be the PEP. The knooppunt can be a PEP without being a proxy.

I'm not against it, but.for now, there's no clear use case or user for that. We can add that as appendix/alteration to this ADR at a later point.

docs/adr/003-knooppunt-solution-architecture.md
```diff
@@ -0,0 +1,132 @@
...
+    - This involves looking up parties, endpoints, authentication
+    - Outbound requests are out of scope for this ADR.
+
+## Considered Options
+This section describes considered architecture options.
+
+In any of the options, there's a proxy (e.g. NGINX, HAProxy, Traefik) in front of the Knooppunt and/or EHR FHIR API.
+This is a typical reverse proxy, handling TLS termination, routing, load balancing, caching, etc.
+
+### Knooppunt as both PEP and PDP
```
stevenvegt:
```suggestion
### Option 1: Knooppunt as both PEP and PDP
```

docs/adr/003-knooppunt-solution-architecture.md
```diff
@@ -0,0 +1,132 @@
...
+
+```text
+┌─────────────────┐       ┌─────────┐       ┌────────────┐       ┌──────────────┐
+│                 │       │         │       │            │       │              │
+│ External System ├──────►│  Proxy  │──────►│ Knooppunt  ├──────►│ EHR FHIR API │
+│                 │       │         │       │ [PEP/PDP]  │       │              │
+└─────────────────┘       └─────────┘       └────────────┘       └──────────────┘
+```
+
+Data exchanges are routed through the Knooppunt, theoretically offloading all "complicated" concerns from the vendor.
```
stevenvegt:
```suggestion
All inbound data requests are routed through the Knooppunt, theoretically offloading all "complicated" concerns from the vendor.
```

docs/adr/003-knooppunt-solution-architecture.md
```diff
@@ -0,0 +1,132 @@
...
+
+- Advantages:
+  - Simpler deployment, since it only requires the Knooppunt to be deployed.
+  - Easier to configure, since all configuration is centralized in the Knooppunt.
+  - Easier to reason about, since all data exchanges go through a single component.
+- Disadvantages:
+  - Large attack surface on the Knooppunt, since it handles all data exchanges.
+    Makes it harder to use security measures vendors already have in place, especially for resource transformation and filtering.
+  - Might not actually make things easier for vendors, if they want to implement requirements not supported by the Knooppunt (e.g. auditing, data minimization)
```
stevenvegt:
```suggestion
- Advantages:
  - Simple deployment, since it only requires the Knooppunt to be deployed.
  - Easy to configure, since all configuration is centralized in the Knooppunt.
  - Easy to reason about, since all data exchanges go through a single component.
- Disadvantages:
  - Large attack surface on the Knooppunt, since it handles all inbound requests.
    Makes it harder to use security measures vendors already have in place, especially for resource transformation and filtering.
  - Less flexible when the vendor needs additional concerns not supported by the Knooppunt (e.g. auditing, data minimization)
```

docs/adr/003-knooppunt-solution-architecture.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Knooppunt as PDP, separate PDP
```
stevenvegt:
```suggestion
### Option 2: Knooppunt as PDP, separate PDP
```

docs/adr/003-knooppunt-solution-architecture.md
```diff
@@ -0,0 +1,132 @@
...
+- Negotiating authentication with the external party
+- Looking up the right endpoint to call
+
+This is not in scope for this ADR.
+
+## Decision Outcome
+
+We have decided to implement the Knooppunt as a Policy Decision Point (PDP), option "Knooppunt as internal system";
+integrating with existing infrastructure via an external proxy for inbound data exchanges.
+This approach was chosen because it minimizes the attack surface of the Knooppunt, leverages existing security components, and provides flexibility for vendors to use their preferred Policy Enforcement Point (PEP).
```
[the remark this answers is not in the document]
reinkrul: According to the reference implementation, that follows XACML, the PDP should retrieve whatever additional info it needs for decisions: https://en.wikipedia.org/wiki/XACML#/media/File:XACML_Architecture_&_Flow.png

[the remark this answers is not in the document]
stevenvegt: Ok, that makes sense. Then I agree with the simplicity of the PEP.

stevenvegt (approved):
We might need an ADR to decide the solution architecture which now sits as a doc with WIP status in the Wiki.

PR #170, opened 2025-09-25
Endpoint Connection Screen

Dirklectisch:
This PR introduces a screen for editing the many to many relationship between organisations and endpoints.

You can reach it by pushing the button on top either edit screen.

Screenshot 2025-09-25 at 13 15 42

The button will lead to a table with checkboxes that let you enable/disable the relationships.

Screenshot 2025-09-25 at 13 18 27

I will address the delete button in a separate PR.

### DIFF
--- a/component/mcsdadmin/component.go
+++ b/component/mcsdadmin/component.go
@@ -78,6 +78,7 @@ func (c Component) RegisterHttpHandlers(mux *http.ServeMux, _ *http.ServeMux) {
 	mux.HandleFunc("GET /mcsdadmin/endpoint", listEndpoints)
 	mux.HandleFunc("GET /mcsdadmin/endpoint/new", newEndpoint)
 	mux.HandleFunc("POST /mcsdadmin/endpoint/new", newEndpointPost)
+	mux.HandleFunc("GET /mcsdadmin/endpoint/connect", connectEndpoint)
 	mux.HandleFunc("GET /mcsdadmin/location", listLocations)
 	mux.HandleFunc("GET /mcsdadmin/location/new", newLocation)
 	mux.HandleFunc("POST /mcsdadmin/location/new", newLocationPost)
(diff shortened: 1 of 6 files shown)


component/mcsdadmin/templates/endpoint_list.html
```diff
@@ -1,7 +1,10 @@
 {{define "main"}}
 <div class="d-flex justify-content-between align-items-center mb-3">
     <h2>Endpoints</h2>
-    <a href="/mcsdadmin/endpoint/new" class="btn btn-primary">New Endpoint</a>
+    <div>
+        <a href="/mcsdadmin/endpoint/connect" class="btn btn-primary">Connect Organizations</a>
```
reinkrul, line 5: I'd call it "Associate" instead of "Connect"

component/mcsdadmin/component.go
```diff
@@ -404,30 +414,134 @@ func newEndpointPost(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, "bad request: could not find organization", http.StatusBadRequest)
 			return
 		}
+
+		var epRef fhir.Reference
+		epRef.Type = to.Ptr("Endpoint")
+		epRef.Reference = to.Ptr("Endpoint/" + *resEp.Id)
+		epRef.Display = to.Ptr(endpoint.Address)
```
reinkrul, line 421: I'd not do this, since if the address changes, it's now out of sync. And it makes people integrate against display, because then they don't have to lookup the endpoint.

reinkrul (changes requested):

reinkrul:
Testing it out, I think this UI is awkward;
Screenshot 2025-09-28 at 09 16 06

reinkrul:
I had something like this in my mind. It doesn't have to be like this, just to show what mean:
Screenshot 2025-09-29 at 08 47 02

Issue #173, opened 2025-09-29
mCSD Update Client: validate resources from care org Admin Directories


reinkrul:
According to the IG;

- Child-resources MUST always refer to an organization (Endpoint, HealthcareService, PractitionerRole, Location)
- Organization-resources MUST always:
  - refer to another Organization resource (if no URA identifier is present)
  - match the Organization resource from the LRZa (if URA identifier is present)

See https://nuts-foundation.github.io/nl-generic-functions-ig/care-services.html#update-client

jorritspee:
Dirk and Bram will review today

PR #174, opened 2025-09-29
#173: validate mCSD updates from mCSD Admin Directories

reinkrul:


component/mcsd/component.go
```diff
@@ -26,6 +26,9 @@ var _ component.Lifecycle = &Component{}
 var rootDirectoryResourceTypes = []string{"Organization", "Endpoint"}
 var defaultDirectoryResourceTypes = []string{"Organization", "Endpoint", "Location", "HealthcareService", "PractitionerRole", "Practitioner"}
 
+// ParentOrganizationMap maps parent organizations (with URA identifier) to their linked child organizations
+type ParentOrganizationMap map[*fhir.Organization][]*fhir.Organization
```
reinkrul: You're only using this type within this package, so no need to export the type (so don't ca[pitaize the first letter of Parent)

component/mcsd/component.go
```diff
@@ -473,3 +485,169 @@ func extractResourceIDFromURL(entry fhir.BundleEntry) string {
...
+	for key, org := range orgMap {
+		if org.Id != nil && *org.Id == orgID {
+			return key
+		}
+	}
+	return ""
+}
+
+// findParentOrganizationWithURA finds a parent organization with a URA identifier and all organizations linked to it.
+// It loops through ALL entries to find organizations with URA identifiers, then selects the one with the most
```
Dirklectisch: Does it really take the amount of child relationships into account? It think it just builds the index independent of the amount of relationships, right?

component/mcsd/component.go
```diff
@@ -473,3 +485,169 @@ func extractResourceIDFromURL(entry fhir.BundleEntry) string {
...
+	// First try to find parent organizations from deduplicatedEntries
+	parentOrganizationsMap, err := findParentOrganizationWithURA(ctx, deduplicatedEntries)
+	if err != nil {
+		return nil, err
+	}
+
+	// If no parent organizations with URA were found, query all organizations from the directory
+	if len(parentOrganizationsMap) == 0 {
+		log.Ctx(ctx).Debug().Str("fhir_server", fhirBaseURLRaw).Msg("No parent organizations with URA found in deduplicatedEntries, querying all organizations")
+		orgEntries, _, err := c.queryHistory(ctx, remoteAdminDirectoryFHIRClient, "Organization", url.Values{
```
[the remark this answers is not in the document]
gasperr: if you are processing a history on Endpoints, for example, some Endpoints may have been referenced on organizations of previous versions. And therefore validation of such Endpoint would fail, even though it was in fact referenced to that Org. That's why I implemented the history check instead of just the regular one.

What may mitigate this (discussed with Bram) is that Organizations rarely change so there shouldn't be that much of a difference between history and actual state. Although that argument may be a thin one when implementing a future proof software, so I am saying this knowing it's a bad take :)

reinkrul:
I reviewed, but appearantly I forgot to click the "submit" button... But I discussed my points with @Dirklectisch , so if he's OK with it, don't wait for me to review.

PR #175, opened 2025-09-30
Deployment documentation

reinkrul:
Adds:
- Deployment documentation overview, with options
- Script for generating diagrams from the C4 file


docs/generate.sh
```diff
@@ -0,0 +1,4 @@
+docker run --rm -v "${PWD}:/docs:ro" -v ./images:/images \
+  extenda/structurizr-to-png \
+  --path c4-diagram.dsl \
```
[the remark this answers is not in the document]
reinkrul: it's not

docs/c4-diagram.dsl
```diff
@@ -0,0 +1,226 @@
...
+                tags "database"
+            }
+
+            fhirQueryDir = container "mCSD Query Directory" "Stores mCSD resources for querying" {
+                tags "database"
+                technology "HAPI FHIR"
+            }
+            fhirAdminDir = container "mCSD Administration Directory" "Stores mCSD resources for synchronization" {
+                tags "database"
+                tags "new-fhir-admin-directory"
```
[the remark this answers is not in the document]
reinkrul: ?

docs/images/structurizr-C2_XIS_ContainerDiagram.png
Dirklectisch, line 1: In this diagram it seems the update client is calling the nuts knooppunt directly, I don't think that's the case unless we made the decision to start proxying the MCSD admin.

It looks in this diagram like the nuts node is part of the XIS which I don't think it can be?

docs/images/structurizr-A2_ContainerDiagram.png
Dirklectisch, line 1: I wonder it's misleading to put the FHIR servers here within the boundaries of the knooppunt system. It might give the impression the vendor doesn't need to do anything to set that up. Should it be more clear that those or managed by the vendor?

Dirklectisch, line 1: Idea: Would it be useful to keep the system boundaries like they are in A1 but now with an additional box your local services and systems?

docs/images/structurizr-B2_ContainerDiagram.png
Dirklectisch, line 1: For this use case it should be super clear there's work to be done to properly expose the admin directory to the outside world. If you are currently on the "one FHIR store to rule them all" setup some thought needs to be put into proxying etc. This diagram makes me feel like you can simply point knooppunt to your existing store and that's it.

Dirklectisch (approved):
I wonder if there is an opportunity to organize the diagrams on a per generic function basis instead of intersections of each variant.

Other than that happy we have some more diagrams now!

PR #195, opened 2025-10-09
Docs: change diagrams to have a set of diagrams per GF

reinkrul:
Our previous diagrams focused on deployment variants, but got quite cluttered. Dirk asked me to change the approach, to focus on GF function instead. So now, we have diagrams of things involved for each generic function.
Note that it reflects the current state of the implementation, so things we're still working on, or we wish/know we'll build aren't there by intention, to keep it crystal clear about what we actually have.


Dirklectisch (approved):
I really like these thanks for making them!

Issue #192, opened 2025-10-08
Fake NVI: validate custodian equals X-Requestor-URA


gasperr:
add a check in the NVI HAPI interceptor that when registering, the X-Requestor-URA matches the custodian's URA

Otherwise, you can register documentreferences for other parties (which is weird), and will probably come up during testing/integration

PR #200, opened 2025-10-14
#192: validate custodian with ura header, if exists

gasperr:


Dirklectisch (approved):
I was wondering if this validation is a security measure or just a sanity check. This still let's you commit document references for other parties as you can simply omit the header and it will pass.

reinkrul:
> > I was wondering if this validation is a security measure or just a sanity check. This still let's you commit document references for other parties as you can simply omit the header and it will pass.
> 
> Yes this is more of a sanity check than anything else. Especially considering whole Organization can be referenced from the DocumentReference, meaning we'd have to resolve it from the source system to really validate URA in there (which would mean backward-facing connection).
> 
> Since this is a "fake nvi" for convenience sake for the time of development/testing and not something we'd be delivering in production, I feel it does what it best can do for the lowest amount of effort.
> 
> But since the issue was raised by @reinkrul, your 2 cents would be appreciated as well.

I agree, and at the same time it shouldn't cause bugs to go unnoticed, e.g. us not sending the requester URA (to fake NVI) at all, causing things not to be pseudonymized. That was the bug (in NVI Gateway) that I ran into, which led me to raise the issue.

With the matching on HAPI tenant to activate this interceptor, every call to it should have an X-Requester-URA, so I think I would've implemented the "X-Requester-URA MUST be set" business rule, but for me it's potato/potato (because, as you said, it's test tooling for now)

gasperr:
I'll do that @reinkrul.. also removes some unnecessary complexity. So I'll just enforce that header with every request. Stand by.

Issue #139, opened 2025-09-12
Add PractitionerRole to mCSD Admin


Dirklectisch:
Make the practitionerRole editable with the following fields.

MUST PractitionerRole.Practitioner: enter DEZI-number (.practitioner.identifier.system and .practitioner.identifier.code)
MUST PractitionerRole.Organization: select existing Organization
MUST PractitionerRole.code: dropdown, select multiple values
MUST PractitionerRole.telecom: multiple system+value entries, system is a dropdown with codes 'phone', 'email' and 'url' (url is meant for matrix-connection details for OZOverbindzorg in PoC 10)

View the full data model here:

<img width="1153" height="584" alt="Image" src="https://github.com/user-attachments/assets/04789cd6-554b-40de-9710-270a09099594" />

PR #201, opened 2025-10-16
#139 Add practitioner role in mcsd admin

Dirklectisch:
This PR makes it possible to add PractitionerRoles through the MCSD Admin application.

This form included options that spanned multiple fields and I've therefore introduced some code that handles field names in this format.

`telecom[0][Value]`

The backend parses this format using some helper functions in the new `formdata` module.

Needed to add some more logic to the frontend to handle these more complex option pairs. Alternative approach would be to render that part in the backend but I figured it would still require some JS in that case to say count the amount of current options filled in.

Additionally this has some cool HTMX stuff around error handling. The server will now figure out based on the headers if the error needs be rendered inline or as a full page and render the appropriate HTML.

### DIFF
--- a/component/mcsdadmin/component.go
+++ b/component/mcsdadmin/component.go
@@ -105,10 +105,10 @@ func listServices(w http.ResponseWriter, _ *http.Request) {
 	renderList[fhir.HealthcareService, tmpls.ServiceListProps](client, w, tmpls.MakeServiceListXsProps)
 }
 
-func newService(w http.ResponseWriter, _ *http.Request) {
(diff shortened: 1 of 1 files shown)


gasperr (approved):

JorisHeadease:
I would actually argue that the form order is safer via HTML (its submission order is guaranteed by the HTML spec) than in Go, the current code uses `map[index]map[key]value{}`, but Go maps do not guarantee order.

The empty field is a valid concern, but the inputs are required, so that would rule out the form being submitted. We could also add a defensive check in the background to verify the count equals or so.

PR #208, opened 2025-10-20
Add ADR4 for consent PDP API

stevenvegt:
This ADR describes the problem of the NVI which needs to know if it can share lokalisation information with the requesting party and suggest a solution of using a PDP at the custodian.


stevenvegt:
Moved to https://github.com/nuts-foundation/nl-generic-functions-ig/issues/42

PR #210, opened 2025-10-23
C4 diagrams for PEP/PDP that use GF Consent

### LINKED ISSUE #217: Update architectural documentation to include to PEP and data exchanges

reinkrul:
Fixes #217


docs/images/structurizr-DataExchange_ContainerDiagram.svg
Dirklectisch: What's the purpose of having both this ContainerDiagram and the ComponentDiagram in this PR? They look very similar to me. Just a slight difference in granularity.

Dirklectisch (approved):
I think maybe it would be useful also link to the SSIBAC reference architecture diagram somewhere. For me that diagram clarified a lot when understand what we are aiming for.

Issue #216, opened 2025-10-24
Implement PDP endpoint in Knooppunt


reinkrul:
Mount on internal API (`:8081`). Implement a hardcoded policy that checks for consent at Mitz

Follow Open Policy Agent's API (you can check https://github.com/nuts-foundation/nuts-pxp for inspiration)

Decisions made:

- The goal of this work item is to create an endpoint that is ready to integrate with for the hackathon next week. Not to make a fully flexible PDP endpoint.
- The endpoint will follow OPA conventions but for this MVP does not aim to use the OPA service or Go Library for its implementation.
- A single endpoint wil be provided under `/pdp` that implements OPA's query API. It will implement only the main policy for now that does the consent check. (example here: https://www.openpolicyagent.org/docs/rest-api#query-api)
- To ease the integration with proxy's we will stay as close to OPA conventions as possible. If we want to provide the Authzen format we can do so separately. (example here: https://www.openpolicyagent.org/docs/http-api-authorization#3-check-that-alice-can-see-her-own-salary )
- The PDP endpoint should be available on the internal port 8081 only.

PR #219, opened 2025-10-27
#216 Boilerplate PDP endpoint

Dirklectisch:
Implements a PDP endpoint according to the OPA query API but always returns false.

### DIFF
--- a/component/pdp/component.go
+++ b/component/pdp/component.go
@@ -96,18 +96,7 @@ func (c Component) HandleMainPolicy(w http.ResponseWriter, r *http.Request) {
 
 	ctx := context.Background()
 	mitzComp := *c.Mitz
-	consentReq := xacml.AuthzRequest{
-		PatientBSN:             input.PatientBSN,
-		HealthcareFacilityType: input.DataHolderFacilityType,
-		AuthorInstitutionID:    input.DataHolderOrganizationUra,
-		// This code is always the same, it's the code for _de gesloten vraag_
-		EventCode:              "GGC002",
-		SubjectRole:            input.RequestingUziRoleCode,
-		ProviderID:             input.RequestingPractitionerIdentifier,
-		ProviderInstitutionID:  input.RequestingOrganizationUra,
-		ConsultingFacilityType: input.RequestingFacilityType,
-		PurposeOfUse:           "TREAT",
-	}
+	consentReq := xacmlFromInput(input)
 	consentResp, err := mitzComp.CheckConsent(ctx, consentReq)
 	if err != nil {
 		http.Error(w, "could not complete the consent check", http.StatusInternalServerError)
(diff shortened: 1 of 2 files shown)


Dirklectisch:
Tested this out using our test Mitz proxy and the .http script in this PR and seems to work alright.

docs/test-scripts/pdp.http
```diff
@@ -0,0 +1,31 @@
+### Testing valid request
+
+POST http://localhost:8081/pdp/v1/data/knooppunt/authz
+Content-Type: application/json
+
+{
+  "input": {
+    "method": "GET",
+    "Path": ["fhir", "Patient", "118876"],
```
JorisHeadease:
```suggestion
    "path": ["fhir", "Patient", "118876"],
```

component/pdp/component.go
```diff
@@ -0,0 +1,180 @@
...
+	consentReq := xacml.AuthzRequest{
+		PatientBSN:             input.PatientBSN,
+		HealthcareFacilityType: input.DataHolderFacilityType,
+		AuthorInstitutionID:    input.DataHolderOrganizationUra,
+		EventCode:              "GGC002",
+		SubjectRole:            input.RequestingUziRoleCode,
+		ProviderID:             input.RequestingUziRoleCode,
+		ProviderInstitutionID:  input.RequestingOrganizationUra,
+		ConsultingFacilityType: input.RequestingFacilityType,
+		PurposeOfUse:           "TREAT",
```
JorisHeadease: We send this as `treatment` from the PEP, we'll probably need a translation here

component/pdp/component.go
```diff
@@ -0,0 +1,180 @@
...
+	Path                             []string `json:"path"`
+	PatientBSN                       string   `json:"patient_bsn"`
+	RequestingUziRoleCode            string   `json:"requesting_uzi_role_code"`
+	RequestingPractitionerIdentifier string   `json:"requesting_practitioner_identifier"`
+	RequestingOrganizationUra        string   `json:"requesting_organization_ura"`
+	RequestingFacilityType           string   `json:"requesting_facility_type"`
+	DataHolderOrganizationUra        string   `json:"data_holder_organization_ura"`
+	DataHolderFacilityType           string   `json:"data_holder_facility_type"`
+	PurposeOfUse                     string   `json:"purpose_of_use"`
+}
```
JorisHeadease: We also send `ResourceType` and `ResourceId` due to OPA. We don't use them, but it might be nice to add them to the struct as they are part of its input.

component/pdp/component.go
```diff
@@ -0,0 +1,180 @@
...
+
+	allow := false
+	switch consentResp.Decision {
+	case xacml.DecisionPermit:
+		allow = true
+	case xacml.DecisionDeny:
+		allow = false
+	default:
+		allow = false
+	}
```
JorisHeadease: This feels a bit elaborate. 

```
allow := consentResp.Decision == xacml.DecisionPermit
```

Or if you prefer being explicit with the if statement:

```
allow := false
if consentResp.Decision == xacml.DecisionPermit {
    allow = true
}
```

JorisHeadease (changes requested):
Added a review. In general, it might also be nice to add some unit tests to the component, especially around mappings from pep towards pdp, so they remain in tact

JorisHeadease (approved):
Looks solid @Dirklectisch!

PR #240, opened 2025-11-04
(PDP) Check for Resource ID Removed

melihaydogd:
This check prevents us doing search requests while requesting data from the PEP. It only allows read requests.

### DIFF
--- a/component/pdp/component.go
+++ b/component/pdp/component.go
@@ -160,7 +160,6 @@ func validateInput(input MainPolicyInput) bool {
 		input.DataHolderOrganizationUra,
 		input.DataHolderFacilityType,
 		input.PurposeOfUse,
-		input.ResourceId,
 		input.ResourceType,
 	}
 	if slices.Contains(requiredValues, "") {


Dirklectisch:
The PDP endpoint is still a little in flux as long as we don't have the implementation guide for authentication done. However I do think wee need to know the resource ID's to figure out who can access what. Could be that we are turning this property into a list rather than a single ID?

Could use some guidance from @reinkrul / @bramwesselo on how to handle search queries.

melihaydogd:
I think a better solution would be to add a `patient.identifier` parameter (or `identifier` for `Patient` request) in the PEP component before forwarding the request to resource server.
Regardless, if we need resource ID to test everyone's PEP, then everyone needs to share their resources' ID. Since this isn't how it is gonna work at the end, having the test like that wouldn't be ideal in my opinion. In our case, I am working on ZNO integration and it wouldn't work if the search requests are not allowed.

Dirklectisch:
If `ResourceId` is required or not will henceforth be determined based on the policies attached to the given scope. In any case the PDP api is in flux while we are further developing it at the moment. So I think we can close this for now.

Issue #257, opened 2025-11-18
Implement CapabilityStatement Policy in PDP endpoint


Dirklectisch:
As detailed in the [PEP/PDP proposal](https://github.com/nuts-foundation/nuts-knooppunt/pull/244/files) we will implement a policy based on capability statements.

- The `scope` property will determine what capability statement is used. This allows for authentication to be done against a certain scope. Example scope can be `eOverdracht-server`.
- The capability statement will be fetched from the PIP point, which is modelled as a FHIR service. For local development we can setup another FHIR tenant in our local HAPI server.
- The capability statement determines what a client can do, for example the MCSD admin update client may only do a `_history` query.
- Part of this ticket will be some research which properties of a capability we want to support for policy enforcement.

PR #267, opened 2025-11-26
#257 Introduce capability checks in PDP

Dirklectisch:
This PR introduces check against capability statements in the PDP.

Theres' still more work to do when it comes to supporting more features of those statements but this covers basic interactions with resources.

### DIFF
--- a/component/pdp/component.go
+++ b/component/pdp/component.go
@@ -98,7 +98,7 @@ func (c Component) HandleMainPolicy(w http.ResponseWriter, r *http.Request) {
 		}
 		writeResp(w, Allow())
 	case "patient_example":
-		writeResp(w, EvalMitzPolicy(c, input))
+		writeResp(w, EvalMitzPolicy(c, r.Context(), input))
 	default:
 		writeResp(w, Deny(
 			ResultReason{
--- a/component/pdp/mitz.go
+++ b/component/pdp/mitz.go
@@ -7,7 +7,7 @@ import (
 	"github.com/nuts-foundation/nuts-knooppunt/component/mitz/xacml"
 )
 
-func EvalMitzPolicy(c Component, input MainPolicyInput) PolicyResult {
+func EvalMitzPolicy(c Component, ctx context.Context, input MainPolicyInput) PolicyResult {
 	// TODO: make this return more detailed information for what fields are missing
 	ok := validateMitzInput(input)
 	if !ok {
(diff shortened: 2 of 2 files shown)


test/testdata/vectors/pip/vectors.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+func HAPITenant() hapi.Tenant {
+	return hapi.Tenant{
+		Name: "policy-information-point",
+		ID:   8,
+	}
+}
+
+func CapabilityStatements() []fhir.CapabilityStatement {
```
reinkrul: maybe we can do a ``//go:generate`` instruction that copies this from `component/pdp/capabilities`, and unmarshal that? Then we don't need to keep it in sync?

component/pdp/capabilities.go
```diff
@@ -0,0 +1,112 @@
...
+		return fhir.CapabilityStatement{}, false
+	}
+}
+
+func EvalCapabilityPolicy(input MainPolicyInput) PolicyResult {
+	out := PolicyResult{
+		Allow: false,
+	}
+
+	if input.Scope == "" {
```
reinkrul: does this function need to be exported? Because you already did this in the calling function, so if you make this private, you don't need to re-check this?

component/pdp/capabilities.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		out.Reasons = []ResultReason{reason}
+		return out
+	}
+
+	statement, ok := capabilityForScope(input.Scope)
+	if !ok {
+		reason := ResultReason{
+			Code:        "unexpected_input",
+			Description: "unexpected input, no capability statement known for scope",
```
reinkrul: do we always want to fail on this? Because I don't think that what was built at the hackaton, has a CapabilityStatement (and it will break if we merge this).

Ultimately, you probably want to have that configurable; which scopes are accepted, and whether (or which) capabilitystatement is authorized.

test/testdata/vectors/pip/vectors.go
reinkrul: Why do we even need this here?

component/pdp/capabilities.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	allowInteraction := false
+	for _, des := range resourceDescriptions {
+		for _, inter := range des.Interaction {
+			if inter.Code == interaction {
+				allowInteraction = true
+			}
+		}
+	}
```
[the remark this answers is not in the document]
Dirklectisch: Changed my mind on this. I've implemented the search params for `searchType` so that we support everything in the MCSD statement.

Made it so that interaction types I haven't implemented yet get rejected so to prevent any mistakes. Will have a look at what is necassary for the BGZ in a next PR.

reinkrul (approved):
minor comment

PR #274, opened 2025-12-04
#257 support for include and revinclude

Dirklectisch:
This PR adds support for revinculde and includes when enforcing the capability statement policy.

Not covered by this PR are system searches, batches/transactions and operations (will implement them next)

### DIFF
--- a/component/pdp/fhirreq.go
+++ b/component/pdp/fhirreq.go
@@ -44,6 +44,11 @@ var definitions = []PathDef{
 		Verb:        "DELETE",
 	},
 	{
+		Interaction: fhir.TypeRestfulInteractionDelete,
+		PathDef:     []string{"[type]?"},
+		Verb:        "DELETE",
+	},
+	{
 		Interaction: fhir.TypeRestfulInteractionCreate,
 		PathDef:     []string{"[type]"},
 		Verb:        "POST",
@@ -52,6 +57,11 @@ var definitions = []PathDef{
 		Interaction: fhir.TypeRestfulInteractionSearchType,
 		PathDef:     []string{"[type]?"},
 		Verb:        "GET",
+	},
+	{
+		Interaction: fhir.TypeRestfulInteractionUpdate,
+		PathDef:     []string{"[type]?"},
+		Verb:        "PUT",
 	},
 	{
 		Interaction: fhir.TypeRestfulInteractionSearchType,


component/pdp/capabilities.go
```diff
@@ -140,21 +135,67 @@ func evalInteraction(
...
+	for _, inc := range input.Include {
+		if !slices.Contains(allowedIncludes, inc) {
+			rejectedIncludes = append(rejectedIncludes, inc)
 		}
+	}
+	if len(rejectedIncludes) == 0 {
+		allowIncludes = true
+	}
+
+	if !allowIncludes {
```
reinkrul: Why use this var?

To me, the following seems simpler:

```golang
if len(rejectedIncludes) > 0 {
   // ...
}
```

component/pdp/component.go
```diff
@@ -92,17 +92,20 @@ func (c Component) HandleMainPolicy(w http.ResponseWriter, r *http.Request) {
...
 			writeResp(w, Deny(ResultReason{
-				Code:        "not_allowed",
+				Code:        TypeResultCodeNotAllowed,
 				Description: "not allowed to request this resources during update",
 			}))
 		}
 		writeResp(w, Allow())
+	case "mcsd_query":
+		// Dummy should be replaced with the actual OPA policy
+		writeResp(w, Allow())
```
reinkrul: Just eval-ing the capabilitystatement should be enough?

component/pdp/capabilities.go
```diff
@@ -140,21 +135,67 @@ func evalInteraction(
...
+		reasons := make([]ResultReason, 0, len(rejectedIncludes))
+		ManyReasons(&reasons, rejectedIncludes, "include %s is not allowed", TypeResultCodeNotAllowed)
```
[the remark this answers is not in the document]
Dirklectisch: Is this version crazy?

You would retain the ability to allocate a slice with whatever capacity and length you prefer, or alternatively the function handles it for you.

````go
func (p *PolicyResult) AddReasons(input []string, format string, code TypeResultCode) {
	isNewSlice := cap(p.Reasons) == 0
	if isNewSlice {
		p.Reasons = make([]ResultReason, len(input))
	}

	for i, str := range input {
		reason := ResultReason{
			Code:        code,
			Description: fmt.Sprintf(format, str),
		}

		if isNewSlice {
			p.Reasons[i] = reason
		} else {
			p.Reasons = append(p.Reasons, reason)
		}
	}
}
````

File: component/pdp/capabilities.go [hunk 4]
```diff
@@ -80,15 +87,11 @@ func evalInteraction(
 		fhir.TypeRestfulInteractionSearchType,
 	}
 	if !slices.Contains(supported, input.InteractionType) {
```
reinkrul: I just noticed that the interaction type is passed by the PEP to the PDP, but I don't think that's correct (at least not in most cases); as that requires the PEP to have in-depth detail about FHIR and do some mapping/parsing (e.g. GET without search params becomes "read", with search params it becomes a "search"). That makes implementing a PEP very complicated.

I think that logic should be in the PDP, making the PEP->PDP request body just user info + HTTP request info (basically URL and query parameters, top of mind), just like when we'd employ OPA + NGINX (for instance).

We might have this FHIR input as intermediate data model for the FHIR policies, but I don't think it should be the API input of the PDP as sent by the PEP.

component/pdp/component.go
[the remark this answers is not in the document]
reinkrul: But there's a capability statement for mCSD Update Client, so why not use that to validate this? Since it also restricts the resources?

component/pdp/shared.go
```diff
@@ -21,6 +26,8 @@ type MainPolicyInput struct {
 	ResourceType                     fhir.ResourceType           `json:"resource_type"`
 	SearchParams                     []string                    `json:"search_params"`
 	ResourceId                       string                      `json:"resource_id"`
+	Include                          []string                    `json:"include"`
+	Revinclude                       []string                    `json:"revinclude"`
```
reinkrul: You need to parse these from the HTTP request URL/method

component/pdp/shared.go
```diff
@@ -58,13 +85,55 @@ func Deny(reason ResultReason) PolicyResult {
...
+	buffer := bytes.Buffer{}
+	enc := json.NewEncoder(&buffer)
+	enc.SetEscapeHTML(false)
+	err := enc.Encode(code.Code())
+	return buffer.Bytes(), err
+}
+func (code TypeResultCode) Code() string {
+	switch code {
+	case TypeResultCodeMissingRequiredValue:
+		return "missing_required_value"
```
reinkrul: If you make the constants type `string`, you don't need custom marshalling?

reinkrul (changes requested):

component/pdp/capabilities.go
```diff
@@ -103,60 +114,84 @@ func evalInteraction(
...
-				{
-					Code:        "not_allowed",
-					Description: "capability statement does not allow interaction",
-				},
-			},
-		}
+		return Deny(
+			ResultReason{
+				Code:        TypeResultCodeNotAllowed,
+				Description: "capability statement does not allow interaction",
```
reinkrul: could multiple CapabilityStatements apply to the scope (or scopes), so at least as one of them gives access, it's fine?

component/pdp/component.go
```diff
@@ -70,48 +69,67 @@ func (c Component) HandleMainPolicy(w http.ResponseWriter, r *http.Request) {
...
+					Code:        TypeResultCodeNotImplemented,
+					Description: "providing multiple qualifications is not yet implemented",
+				},
+			},
+		}
+		writeResp(r.Context(), w, res)
+		return
+	}
+
+	scope := qualifications[0]
```
reinkrul: add TODO

Also (but maybe not too important now), I personally think `scope` should be a first-class citizen in the PDP request? Since that's basically what you're granted access to? (result of "client qualifications").

component/pdp/capabilities.go
```diff
@@ -37,23 +37,31 @@ func capabilityForScope(ctx context.Context, scope string) (fhir.CapabilityState
...
 		return capa, err == nil
-	case "patient_example":
-		capa, err := readCapability(ctx, "patient-example")
+	case "mcsd_query":
+		capa, err := readCapability(ctx, "nl-gf-query-directory-query-client")
+		return capa, err == nil
+	case "bgz_patient":
+		capa, err := readCapability(ctx, "bgz-patient")
+		return capa, err == nil
+	case "bgz_professional":
```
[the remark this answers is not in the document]
Dirklectisch: There should be a system for configuring scopes, their names, their capability statements and rego policies. Do you want to take that as part of this PR?

component/pdp/shared.go
```diff
@@ -2,36 +2,90 @@ package pdp
...
+	Operation       *string                     `json:"operation"`
+	SearchParams    []string                    `json:"search_params"`
+	Include         []string                    `json:"include"`
+	Revinclude      []string                    `json:"revinclude"`
+}
+
+type PolicyContext struct {
+	DataHolderOrganizationId string `json:"data_holder_organization_id"`
+	DataHolderFacilityType   string `json:"data_holder_facility_type"`
+	PatientBSN               string `json:"patient_bsn"`
```
reinkrul: can you make a ticket that we need to decide how to derive this? As we decided earlier to work with patient references (`Patient/<id>`) instead of BSNs

reinkrul (changes requested):

PR #304, opened 2025-12-18
#303: fhirbaseurl of admin directory was treated as key, resulting in…

### LINKED ISSUE #303: Investigate whether knoppunt can sync from 2 separate admin directories with the same url
Investigate whether knoppunt can sync from 2 separate admin directories with the same url. Hypothesis is that it simply ignores other organization, if it has the same admin url.

gasperr:
… admin directories with same fhirbaseurl being skipped from sync

closes #303


Dirklectisch (approved):
Checked out the branch and can confirm I'm now receiving the organisation that wis missing during the demo. Would it make sense to add a test case for this?

gasperr:
> Checked out the branch and can confirm I'm now receiving the organisation that wis missing during the demo. Would it make sense to add a test case for this?

For sure. I'll add a test.

Issue #319, opened 2026-01-08
Fetch Mitz consent record as input for policies when applicable


Dirklectisch:
When checking a policy one might need a consent record based on the information in Mitz.

We need:

- A way to configure the policy information point / EHR endpoints
- A way to determine if the request relates directly to a patient record
- A mechanism to fetch the patient record, determine the BSN and execute the Mitz call
- Translate the Mitz response to a FHIR consent record as input to the policy

PR #320, opened 2026-01-08
#319 Parse patient id

Dirklectisch:
Small PR that parses the patient id so that we can use it down the line to request the patient record from a PIP.


Dirklectisch:
Forgot that the current capability statements need adjustment but added that just now as well. Can we get this reviewed? I'll tackle doing the call to the EHR/PIP for exchanging the PatientId for a BSN next.

PR #322, opened 2026-01-13
#319 PDP: use PIP to fetch BSN and use it in Mitz auth check

Dirklectisch:
This PR let's you configure a PIP for fetching a patient record and determining the BSN for the Mitz check.

### DIFF
--- a/component/pdp/shared.go
+++ b/component/pdp/shared.go
@@ -151,7 +151,7 @@ const (
 )
 
 type PIPConfig struct {
-	URL string
+	URL string `koanf:"url"`
 }
 
 type Config struct {


component/pdp/component.go
```diff
@@ -6,26 +6,45 @@ import (
...
 	"github.com/nuts-foundation/nuts-knooppunt/component"
 	"github.com/nuts-foundation/nuts-knooppunt/component/mitz"
+	"github.com/nuts-foundation/nuts-knooppunt/component/tracing"
 	"github.com/nuts-foundation/nuts-knooppunt/lib/logging"
 )
 
 func DefaultConfig() Config {
 	return Config{
 		Enabled: true,
+		PIPURL:  "",
```
reinkrul: it'll become `KNPT_PIPURL`.

Assuming we want to add things like auth credentials (username password/client certificate/token/oauth2) at some point, it makes sense i.m.o. to make a struct for this. And I don't think we should have "PIP" in the name, because it probably won't resonate with the people using our software. Maybe `KPT_FHIR_URL`?

component/pdp/pip.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+
+		bsns := fhirutil.FilterIdentifiersBySystem(patient.Identifier, coding.BSNNamingSystem)
+		if len(bsns) == 0 {
+			slog.Warn("Could not find BSN for patient record")
+			return policyInput
+		}
+
+		if len(bsns) > 1 {
+			slog.Warn("Patient record has multiple BSN's, defaulting to the first one")
```
reinkrul: I don't think this would be right, maybe error out for now. Because we can't assume the first one is the right one.

component/pdp/component.go
Dirklectisch: Do we want to distinguish between the PIP and the resource server or are they one and the same thing? I could see some use to the information point not being the same as the resource server that is behind the enforcement point.

component/pdp/mitz.go
```diff
@@ -56,7 +53,7 @@ func xacmlFromInput(input PolicyInput) xacml.AuthzRequest {
 	}
 }
 
-func validateMitzInput(input PolicyInput) bool {
+func validateMitzInput(input PolicyInput) PolicyResult {
 	requiredValues := []string{
```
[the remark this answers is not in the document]
Dirklectisch: Guess I did too much [Struct of Array](https://en.wikipedia.org/wiki/AoS_and_SoA) style programming lately. 😉

component/pdp/shared.go
```diff
@@ -147,11 +153,17 @@ const (
...
 )
 
+type PIPConfig struct {
+	URL string
+}
+
 type Config struct {
-	Enabled bool
+	Enabled bool      `koanf:"enabled"`
+	PIP     PIPConfig `koanf:"pipurl"`
```
reinkrul:
```suggestion
	PIP     PIPConfig `koanf:"pip"`
```

component/pdp/shared.go
```diff
@@ -147,11 +153,17 @@ const (
 	TypeResultCodeInternalError        TypeResultCode = "internal_error"
 )
 
+type PIPConfig struct {
+	URL string
```
reinkrul:
```suggestion
	URL string `koanf:"url"`
```

reinkrul (changes requested):

component/pdp/pip.go
```diff
@@ -0,0 +1,52 @@
...
+	"log/slog"
+
+	"github.com/nuts-foundation/nuts-knooppunt/lib/coding"
+	"github.com/nuts-foundation/nuts-knooppunt/lib/fhirutil"
+	"github.com/nuts-foundation/nuts-knooppunt/lib/logging"
+	"github.com/zorgbijjou/golang-fhir-models/fhir-models/fhir"
+)
+
+func PipPolicyInput(c Component, policyInput PolicyInput) PolicyInput {
+	if c.pipClient == nil {
```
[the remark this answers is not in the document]
Dirklectisch: Should be in the policy in my opinion because not all policies (like the MCSD one) require a patient and BSN. So the logic is like this:

- If you provided a patientID check if we can fetch a BSN
- If we have a BSN check if we have consent
- Provide patientId, BSN and consent check to policy to make a decision

reinkrul (approved):

PR #361, opened 2026-01-29
#355 support policies that are not FHIR specific

### LINKED ISSUE #355: Only check capability statement if content type matches
To support not FHIR calls during authorization we need to make the capability statement check conditional.

- Check only if the content type is `application/fhir+json`
- Provide result in policyInput
- Don't reject outright when check fails

To refine: How do we handle non FHIR calls in the PDP?

Dirklectisch:
Closes #355

### DIFF
--- a/component/pdp/component.go
+++ b/component/pdp/component.go
@@ -157,14 +157,13 @@ func (c *Component) HandleMainPolicy(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	regoPolicyResult := *regoPolicyResultPtr
-	if !regoPolicyResult.Allow {
-		notAllowedResult := appendReasons(regoPolicyResult, capabilityResult, mitzResult)
+	if !(*regoPolicyResultPtr).Allow {
+		notAllowedResult := appendReasons(*regoPolicyResultPtr, capabilityResult, mitzResult)
 		writeResp(r.Context(), w, notAllowedResult)
 		return
 	}
 
-	writeResp(r.Context(), w, regoPolicyResult)
+	writeResp(r.Context(), w, *regoPolicyResultPtr)
 }
 
 func writeResp(ctx context.Context, w http.ResponseWriter, result PolicyResult) {


component/pdp/capabilities.go
```diff
@@ -54,10 +58,12 @@ func evalCapabilityPolicy(ctx context.Context, input PolicyInput) PolicyResult {
 			Description: "unexpected input, no capability statement known for scope",
 		}
 		out.Reasons = []ResultReason{reason}
-		return out
+		return input, out
 	}
 
-	return evalInteraction(statement, input)
+	result := evalInteraction(statement, input)
+	input.Context.FhirCapabilityChecked = result.Allow
```
reinkrul: FhirCapabilityChecked -> FHIRCapabilityChecked (by Go convention)

component/pdp/fhirreq.go
```diff
@@ -334,10 +334,19 @@ func derivePatientId(tokens Tokens, queryParams url.Values) (string, error) {
...
+
+	contentType := request.Input.Request.Header.Get("Content-Type")
+	policyInput.Action.Properties.ContentType = contentType
+
 	tokens, ok := parseRequestPath(request.Input.Request)
 	if !ok {
-		reason := ResultReason{Code: TypeResultCodeUnexpectedInput, Description: "Not a valid FHIR request path"}
-		return PolicyInput{}, Deny(reason)
+		// This is not a FHIR request
+		return policyInput, Allow()
```
reinkrul: Maybe merge for now, so we can work in parallel, but I think we need to refactor this further, checking request type here and parsing it (`parseRequestPath()` would become `parseFHIRRequest()` e.g.)

reinkrul (changes requested):

component/pdp/capabilities.go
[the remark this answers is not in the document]
Dirklectisch: Where would you use a more complicated data structure? A simple boolean makes it easy to check in the rego policy.

What we could do is check against multiple capability statements and register a map of name -> boolean results. Is that what you mean?

reinkrul (approved):

Issue #364, opened 2026-02-02
Use connection-type to determine what kind of parsing needs to be done on the request data


Dirklectisch:
For the PDP we should change our approach to determining how request data is parsed.

- Instead of using the content-type header we should be passing the connection-type of the endpoint to the PDP
- The connection type determines what extra parsing should be done to fill out the `PolicyInput`
- There should be a distinction between data coming from the parser and data that assumed to be there for each incoming request

PR #367, opened 2026-02-04
#364 PDP: Use connection type to determine what parsing is done

Dirklectisch:
This PR makes it possible to parse multiple different connection types in the future. It also uses the connection type to determine if we handling a FHIR request in the first places.

### DIFF
--- a/docs/test-scripts/pdp.http
+++ b/docs/test-scripts/pdp.http
@@ -135,3 +135,38 @@ Content-Type: application/json
     }
   }
 }
+
+### Testing pzp call
+POST http://localhost:8081/pdp/v1/data/knooppunt/authz
+Content-Type: application/json
+
+{
+  "input": {
+    "subject": {
+      "properties": {
+        "subject_id": "900030757",
+        "subject_role": "01.015",
+        "subject_organization_id": "00000025",
+        "subject_facility_type": "V5",
+        "client_qualifications": [
+          "pzp_gf"
+        ]
+      }
+    },
+    "request": {
+      "method": "POST",
+      "protocol": "HTTP/1.0",
+      "path": "/Patient/_search?",
+      "header": {
+        "Content-Type": ["application/x-www-form-urlencoded"]
+      },
+      "body": "identifier=http://fhir.nl/fhir/NamingSystem/bsn|900214181"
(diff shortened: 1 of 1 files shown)


component/pdp/policies/pzp/policy.rego
```diff
@@ -17,19 +17,19 @@ allow if {
...
     input.resource.type == "Consent"
-    input.action.properties.interaction_type == "search-type"
+    input.action.properties.connection_data.fhir_rest.interaction_type == "search-type"
     # patient: reference Patient resource
-    startswith(input.action.properties.search_params.patient, "Patient/")
+    startswith(input.action.properties.connection_data.fhir_rest.search_params.patient, "Patient/")
     # _profile
-    is_string(input.action.properties.search_params._profile)
-    input.action.properties.search_params._profile == "http://nictiz.nl/fhir/StructureDefinition/nl-core-TreatmentDirective2"
+    is_string(input.action.properties.connection_data.fhir_rest.search_params._profile)
```
reinkrul: I don't think we need to include the "rest" part in "FHIR", since JSON/XML FHIR are (or should be) semantically the same.

component/pdp/shared.go
```diff
@@ -42,8 +42,9 @@ type HTTPRequest struct {
 }
 
 type PDPContext struct {
-	DataHolderOrganizationId string `json:"data_holder_organization_id"`
+	ConnectionTypeCode       string `json:"connection_type_code"`
 	DataHolderFacilityType   string `json:"data_holder_facility_type"`
+	DataHolderOrganizationId string `json:"data_holder_organization_id"`
 	PatientBSN               string `json:"patient_bsn"`
```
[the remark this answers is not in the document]
Dirklectisch: Hmm yeah I think the thinking is that the patient id is purely parsed from the fhir request while the BSN can come from many places.

component/pdp/policies/pzp/policy.rego
Dirklectisch: The connection type as defined in the addressing function is `hl7-fhir-rest`. There are more fhir connection types like `hl7-fhir-msg` which could have different fields parsed.

PR #375, opened 2026-02-06
PDP: Decode HTTP request parameters, normalize scope (dash-to-underscore)

reinkrul:
This allows PEP to pass in URL-encoded HTTP request body (form encoded) or query parameters (like the encoded `|`), making integration easier.

It also replaces dash (`-`) with an underscore, allowing for scopes with dashes; OPA doesn't support dashes in rules/packages.

### DIFF
--- a/component/pdp/component_test.go
+++ b/component/pdp/component_test.go
@@ -203,6 +203,12 @@ func TestHandleMainPolicy_Integration(t *testing.T) {
 				decision:             true,
 			},
 			{
+				name:                 "allow - patient identifier is encoded",
+				clientQualifications: []string{"pzp-gf"},
+				httpRequest:          `GET /Patient?identifier=http://fhir.nl/fhir/NamingSystem/bsn%7C123456789`,
+				decision:             true,
+			},
+			{
 				name:                 "allow - Patient search with BSN identifier",
 				clientQualifications: []string{"pzp_gf"},
 				httpRequest:          `GET /Patient?identifier=http://fhir.nl/fhir/NamingSystem/bsn|123456789`,


component/pdp/component_test.go
```diff
@@ -196,6 +196,12 @@ func TestHandleMainPolicy_Integration(t *testing.T) {
 	})
 	t.Run("pzp", func(t *testing.T) {
 		testCases := []testCase{
+			{
+				name:                 "allow - dash is normalized to underscore",
+				clientQualifications: []string{"pzp-gf"},
+				httpRequest:          `GET /Patient?identifier=http://fhir.nl/fhir/NamingSystem/bsn|123456789`,
```
Dirklectisch: Should the `|` pipe here be percent encoded now?

component/pdp/fhirreq.go
```diff
@@ -418,16 +434,20 @@ func NewPolicyInput(request PDPRequest) (PolicyInput, PolicyResult) {
...
-				Description: "Could not parse form encoded data",
+				Description: fmt.Sprintf("Could not parse form encoded request body: %v", err),
 			}
 			return PolicyInput{}, Deny(reason)
 		}
-		rawParams = values
+		rawParams = *decodedValues
 	} else {
-		rawParams = request.Input.Request.QueryParams
+		rawParams = policyInput.Action.Request.QueryParams
```
Dirklectisch: Isn't `policyInput.Action.Request.QueryParams` always empty here? If the params are not in the body they are in the PDPInput right not the PolicyInput.

Dirklectisch (approved):

component/pdp/fhirreq.go
gasperr: they're set above in line 388 after being decoded so I think it's good

PR #379, opened 2026-02-06
#376: remove question mark for search paths

### LINKED ISSUE #376: PDP: Fix handling of question mark in paths
Currently, the FHIR request parser matches on question mark (`?`) being in path or not, to determine whether it's a search interaction. This, however, is incorrect. The following cases are search interactions:

- POST on `/_search` or `[type]/_search` (take search params from POST body)
- GET on `/[type]` (take search params from URL query params)

Whether or not a question mark is in the path, should be irrelevant; for the `GET`, no question mark means "no search params specified".

Agree, @gasperr ?

reinkrul:
Closes #376


component/pdp/fhirreq.go
```diff
@@ -57,27 +57,22 @@ var definitions = []PathDef{
...
-	},
 	{
 		Interaction: fhir.TypeRestfulInteractionSearchType,
 		PathDef:     []string{"[type]", "_search"},
 		Verb:        "POST",
 	},
 	{
 		Interaction: fhir.TypeRestfulInteractionSearchSystem,
-		PathDef:     []string{"?"},
+		PathDef:     []string{""},
```
Dirklectisch: This feels a bit inconsistent with the others cases, shouldn't this just be `[]string{}`? In any case it would be good to have a test case for this one as it has a slightly different shape from the others.

reinkrul: I'll add it as sanity check on startup that all path definitions are sane

PR #388, opened 2026-02-11
Unmarshal HealthcareServices used for validation once

phooijenga:
While debugging the mCSD Update Client I noticed that the list of healthcare services used for validation is passed around as a bundle entry, which means it must be unmarshalled before every use.


reinkrul (approved):
I'm not sure whether this solves an actual performance issue, but as it actually lower complexity, I like it 👍

PR #397, opened 2026-02-15
#396: adding pagination to mcsdadmin app

### LINKED ISSUE #396: mCSD Admin app shows max 20 entries in the list, extend this list with pagination
mCSD Admin app shows max 20 entries in the list, extend this list with pagination

gasperr:
Fixes #396

### DIFF
--- a/component/mcsdadmin/templates/templates.go
+++ b/component/mcsdadmin/templates/templates.go
@@ -39,7 +39,7 @@ func RenderWithBase(w io.Writer, name string, data any) {
 	}
 	files = append(files, partialTemplates...)
 
-	ts, err := template.New("base").ParseFS(tmplFS, files...)
+	ts, err := template.ParseFS(tmplFS, files...)
 	if err != nil {
 		slog.Error("Failed to parse template", logging.Error(err))
 		return


component/mcsdadmin/templates/templates.go
```diff
@@ -39,7 +39,28 @@ func RenderWithBase(w io.Writer, name string, data any) {
...
+		"add": func(a, b int) int { return a + b },
+		"sub": func(a, b int) int { return a - b },
+		"mul": func(a, b int) int { return a * b },
+		"min": func(a, b int) int {
+			if a < b {
+				return a
+			}
+			return b
+		},
+		"iterate": func(count int) []int {
```
reinkrul: wouldn this be more of a `range` function?

gasperr:
> I wonder if the infinite scroll pattern described in the HTMX docs would be simpler than the custom function approach? It looks very simple on the surface at least. You simply render the last element to get the next page automatically.
> 
> https://htmx.org/examples/infinite-scroll/

I personally prefer fixed pages and navigation through them and ability to see how many pages/resources there are right away, especially in an admin-like tool (whereas for a social feed this would make sense). Having said that, I'm not against rewriting it to that as well, especially considering it'd lower the complexity and remove that need for that function that @reinkrul also mentioned above.

So.. do we rewrite this current approach, even though it works? Or is it still good enough to merge, because it does address imo quite a critical thing (currently we're limiting our view to 20 resources)?

PR #415, opened 2026-03-02
PDP: Move FHIR Capability Statement check to Rego

reinkrul:
We could then define the FHIR Capability Statement Rego rule in the AuthZ IG.

Pros of having it in Rego (in IG):
- AuthZ implementors using Open Policy Agent, but not the Knooppunt, don't need to implement the check in their own software. This leads to:
  - Safer software (we have 1 audited/tested/checked definition in the IG instead of 10 implementations)
  - Cheaper/simpler implementation (less things to implement)
 
Cons of having it in Rego:
- Somewhat harder to maintain (I'm better at Golang than at Rego)

TODO:
- [ ] Unit tests for the `fhir` package helper rules


reinkrul:
We decided not to do this, because the goal of GF AuthZ policies is to allow configuration of what differs per policy.
Regarding FHIR requests, what differs is the FHIR CapabilityStatement, not how you check them. Meaning:

- FHIR CapabilityStatement is part of the policy
- Checking the FHIR CapabilityStatement v.s. the HTTP request is part of the PDP (not Rego/OPA)

PR #427, opened 2026-03-12
383: policy can have AND and OR search params defined

gasperr:


reinkrul (changes requested):
The issue also states

> We might also want to just let the PEP pass the query string as input to the PDP instead of requiring to do the parsing; this further complicates the parsing, which we should rather do properly in 1 spot, instead of several (and we're already parsing the search query input if it's a POST search).

What do you think, would that make the PEP/proxy simpler? And if so, make it part of this PR, or separate PR?

Since we also have to update the example NGINX PEP in `/pep` (and if we don't change the query string passing right now, we might have to update it twice)

gasperr:
I resolved the conflicts. Can we merge? Migration guide for vendors as follows (where should we put it?):

# Migration Guide

## Branch `383_handling-of-combining-search-params-in-pdp`

### Breaking Change 1: `query_params` replaced by `query`

The `query_params` field in the PDP request JSON body has been replaced by a single `query` string field.

**Before:**
```json
{
  "input": {
    "request": {
      "method": "GET",
      "path": "/Patient",
      "query_params": {
        "_id": ["123"],
        "status": ["active"]
      }
    }
  }
}
```

**After:**
```json
{
  "input": {
    "request": {
      "method": "GET",
      "path": "/Patient",
      "query": "_id=123&status=active"
    }
  }
}
```

Send the raw, percent-encoded query string (everything after `?`) as a plain string in `query`. Do not URL-decode it ? the PDP handles decoding internally. Do not pre-parse it into a key/value map.

reinkrul (approved):

reinkrul:
Maybe we can reach out personally?

PR #456, opened 2026-03-26
Create PIP call for getting resource contents

### LINKED ISSUE #441: Create PIP call for getting resource contents
When the PDP authorises a read interaction it would be useful to have the resource content available for additional checks.

For example if we want to know if you can write to a task to move it to completion we should be able to check if this task was assigned to the authenticated organisation.

JorisHeadease:
- Add `enrichResourceContent` to PIP enrichment pipeline: fetches the targeted resource (type + id) from PIP and populates `resource.content` in the policy input for OPA policies to inspect
- Best-effort: returns `pip_error` reason when resource can't be fetched, but doesn't block policy evaluation
- Skips enrichment for requests without a specific resource (searches, creates without id)
- Fix pre-existing bug in `enrichConsent` where the scope coding condition could panic on 0 codings or fall through incorrectly

Closes #441


component/pdp/shared.go
```diff
@@ -174,6 +174,7 @@ type PolicyResource struct {
 	Id        string             `json:"id"`
 	Type      *fhir.ResourceType `json:"type"`
 	VersionId string             `json:"version_id"`
+	Content   map[string]any     `json:"content,omitempty"`
```
reinkrul, line 177: I was thinking that it could be anything, not just `map[string]any`. But it doesn't matter much right now, as we only supported structured content (FHIR) for now. So it's fine 🙃

component/pdp/pip.go
```diff
@@ -84,6 +86,36 @@ func (c *Component) enrichBSN(ctx context.Context, policyInput *PolicyInput) (*P
...
+	ctx, cancel := context.WithTimeout(ctx, pipTimeout)
+	defer cancel()
+
+	client := c.pipClient
+	path := fmt.Sprintf("%s/%s", policyInput.Resource.Type.String(), policyInput.Resource.Id)
+
+	var content map[string]any
+	err := client.ReadWithContext(ctx, path, &content)
+	if err != nil {
+		slog.WarnContext(ctx, "Failed to get resource content from PIP", logging.Error(err))
```
reinkrul: should we add the path/URL that was attempted? (I think `client` has a `URL()` function, or if it doesn't at least log the reference)

component/pdp/pip.go
```diff
@@ -84,6 +86,36 @@ func (c *Component) enrichBSN(ctx context.Context, policyInput *PolicyInput) (*P
...
+
+func (c *Component) enrichResourceContent(ctx context.Context, policyInput *PolicyInput) (*PolicyInput, []ResultReason) {
+	if policyInput.Resource.Type == nil || policyInput.Resource.Id == "" {
+		return policyInput, nil
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, pipTimeout)
+	defer cancel()
+
+	client := c.pipClient
```
reinkrul: I don't think 1 usage justifies an extra variable just because it's _somewhat_ shorter to write

reinkrul (changes requested):

component/pdp/shared.go
JorisHeadease: Yeah, it's currently always sent to `to.JSONMap()`, we can make it broader when needed

reinkrul:
@Dirklectisch I think the vendors currently using the PDP don't actively support responding to this query. There's also no policies yet that make use of this. Should we put this behind a feature/config flag, until it's properly supported and used, to prevent causing lots of error/warning logging?

Dirklectisch:
Sure makes sense to me to leave this as an optional advanced feature for now.

reinkrul (approved):

PR #457, opened 2026-03-26
Extend e-overdracht policy

Dirklectisch:
- Add local consent check for task updates
- Allow document operation on composition


reinkrul (approved):
Minor comments

PR #463, opened 2026-03-30
Add OPA policy tests for all 7 PDP policies

### LINKED ISSUE #458: PDP: Data-driven test suites for policies
We envision a policy repository with high-quality policies. A policy can only be trustworthy and high-quality if there's sufficient test cases.

Our current policy tests are in the form of unit tests in Golang, but they are neither sufficient (coverage), readable nor very scalable. Maybe we can employ the official way of testing policies to improve on that? (https://www.openpolicyagent.org/docs/policy-testing)

Note that it can even provide code coverage.

JorisHeadease:
## Summary
- Add 80 OPA/Rego tests across all 7 PDP policies achieving 99.8% coverage
- Integrate `opa test` with coverage reporting and `opa check --strict` into CI
- Coverage percentage reported in GitHub Actions job summary

Closes #458


.github/workflows/run-tests.yaml
```diff
@@ -12,7 +12,22 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    
+
+    - name: Install OPA
```
reinkrul: I think we generally call it setting up; e.g. "Set up OPA"

reinkrul (changes requested):
Can you slim down `TestHandleMainPolicy_Integration()`, which contains policy-specific tests? They are now duplicate with these tests.

reinkrul (approved):

PR #471, opened 2026-04-08
Documentation on the nvi integration for the additional header

gasperr:


docs/INTEGRATION.md
```diff
@@ -146,6 +146,21 @@ pseudo:
...
 
+### Required request header: X-Tenant-ID
+
+Every request to the NVI endpoints must include the `X-Tenant-ID` header identifying your organization:
+
+```http
+X-Tenant-ID: http://fhir.nl/fhir/NamingSystem/ura|<your-ura>
+```
+
+This header is required because BSN pseudonymization is tenant-specific: the Knooppunt uses your URA to generate
```
reinkrul, line 157: and it is also used for the NVI knowing who's registering the record

gasperr: no that's present in the ListResource itself for the NVI. Header itself isn't propagated nor required for NVI.

reinkrul: ah right, but X-Tenant-ID must then match the URA contained in the ListResource? Isn't that duplicate information then? Should we deprecate X-Tenant-ID, or keep it (and check they're the same to prevent accidents).

gasperr: We can't deprecate it, because we need that information for a GET interactions. But we could make additional validation that it matches that withint he ListResource (or even auto populate it in the ListResource when creating an entry).

reinkrul: What do you think?

reinkrul (approved):

docs/INTEGRATION.md
reinkrul: (approving as its out of scope for this PR)

gasperr: Yeah I will make it so. #474

Issue #474, opened 2026-04-08
Autopopulate ListResource from X-Tenant-Id header in NVI integration


gasperr:
Autopopulate ListResource from X-Tenant-Id header in NVI integration and make sure the two values match if both are present.

gasperr:
When posting a new NVI entry (via POST /nvi bundle or POST /nvi/List), the X-Tenant-ID header value (URA) is now reconciled with the nl-gf-localization-custodian extension on the List resource:

If the extension is absent, it is automatically populated from the tenant ID header before forwarding to NVI.
If the extension is present and matches the header, the request proceeds as normal.
If the extension is present but mismatches the header, a 400 Bad Request is returned with a descriptive error.

#479

PR #479, opened 2026-04-13
#474: NVI X-Tenant-ID header to match List custodian

gasperr:
When posting a new NVI entry (via POST /nvi bundle or POST /nvi/List), the X-Tenant-ID header value (URA) is now reconciled with the nl-gf-localization-custodian extension on the List resource:                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                         
  - If the extension is absent, it is automatically populated from the tenant ID header before forwarding to NVI.                                                                                                                                                                                                                                                                                                        
  - If the extension is present and matches the header, the request proceeds as normal.                                                                                                                                                                                                                                                                                                                                
  - If the extension is present but mismatches the header, a 400 Bad Request is returned with a descriptive error.


component/nvi/component.go
```diff
@@ -414,6 +417,37 @@ func (c Component) tokenizeListIdentifiers(ctx context.Context, resource fhir.Li
...
+		if *ext.ValueReference.Identifier.Value != localOrganizationURA {
+			return fhirapi.BadRequestError(
+				fmt.Sprintf("custodian extension URA (%s) does not match tenant ID header (%s)", *ext.ValueReference.Identifier.Value, localOrganizationURA),
+				nil,
+			)
+		}
+		return nil
+	}
+	// Extension absent or incomplete — add/replace it.
+	resource.Extension = append(resource.Extension, fhir.Extension{
```
reinkrul, line 440: this doesn't replace it, but only add it?

reinkrul (changes requested):

reinkrul (approved):

PR #483, opened 2026-04-16
feat: mask BSNs in OPA decision logs and MITZ logs

### LINKED ISSUE #349: Mask BSNs in OPA decision logs
Currently, OPA logs the whole input as part of the decision logs. The input often contains BSNs, so this most probably isn't OK.

See https://www.openpolicyagent.org/docs/management-decision-logs#masking-sensitive-data

Slack msg;
> OPA currently logs BSNs as part of its decision logs. The decision logs are written to our general logs, making it a problem (logging BSNs is generally not OK due to data protection law). We have 2 options:Do not log decision logs as "regular" logging, but allow it to be sent somewhere else. One could argue it should be part of the audit logging?Mask the BSNs in the decision logs (the quick option).What do you think? My proposal would be option 2 for now, audit logging later if needed.

JorisHeadease:
Closes #349.

The issue was scoped to OPA decision logs. While investigating I found MITZ XACML debug logs leaked BSNs in the same way, so I expanded the scope to cover both.

**OPA decision logs** (new `data.system.log.mask` rego rule)
- Redacts `context.patient_bsn`, `subject.patient_enrollment_identifier`, and `search_params.identifier` (URI and OID forms of the BSN NamingSystem, case-insensitive)
- Debug-only fields `request.query`, `request.body`, `request.header`, `resource.content` replaced with `[REDACTED]` placeholder so the presence of the field remains visible for triage
- `is_string` guard on mask helpers prevents fail-open on unexpected input shapes
- `system` bundle blocked from being invoked as a policy via scope
- E2E test builds a fully-populated `PolicyInput` struct and asserts the captured slog stream from a real OPA decision contains `[REDACTED]` but not the BSN, with a guard against vacuous pass

**MITZ XACML logs** (found during investigation, same root cause)
- Request, response and error bodies scrubbed via OID-matching regex on the HL7 `InstanceIdentifier`. Catches BSNs echoed back by MITZ that were never sent in the original request, such as those of a legal representative.

### DIFF
--- a/component/mitz/component.go
+++ b/component/mitz/component.go
@@ -151,10 +151,10 @@ func (c *Component) CheckConsent(ctx context.Context, authzReq xacml.AuthzReques
 		return nil, fmt.Errorf("failed to create authorization decision query: %w", err)
 	}
 
-	// Log the XML request
+	// Log the XML request with every BSN-scoped InstanceIdentifier redacted.
 	slog.DebugContext(ctx, "Sending consent check request to MITZ",
 		slog.String("endpoint", c.consentCheckEndpoint),
-		slog.String("xmlPayload", authnDecisionQueryXml))
+		slog.String("xmlPayload", xacml.RedactBSN(authnDecisionQueryXml)))
 
 	// Create HTTP request with XML payload
 	httpReq, err := http.NewRequestWithContext(
@@ -185,14 +185,19 @@ func (c *Component) CheckConsent(ctx context.Context, authzReq xacml.AuthzReques
 		return nil, fmt.Errorf("failed to read consent check response: %w", err)
 	}
 
-	// Log response
+	// Redact every BSN-scoped InstanceIdentifier before the body touches a log
+	// or an error chain. MITZ echoes the resource-id (BSN) back in the response
+	// because IncludeInResult="true" is set on the request, and may also include
+	// BSNs we never sent (e.g. of a legal representative).
+	redactedBody := xacml.RedactBSN(string(responseBody))
+
 	slog.DebugContext(ctx, "Received consent check response from MITZ",
 		slog.Int("statusCode", resp.StatusCode),
-		slog.String("responseBody", string(responseBody)))
+		slog.String("responseBody", redactedBody))
 
 	// Check for non-2xx status
 	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
-		return nil, fmt.Errorf("consent check failed with status %d: %s", resp.StatusCode, string(responseBody))
+		return nil, fmt.Errorf("consent check failed with status %d: %s", resp.StatusCode, redactedBody)
 	}
 
 	// Parse the XACML response to extract the decision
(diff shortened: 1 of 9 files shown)


component/mitz/xacml/redact.go
```diff
@@ -0,0 +1,35 @@
...
+const redactedExtension = `extension="[REDACTED]"`
+
+// Two patterns cover both attribute orderings XACML emitters may produce.
+// They match InstanceIdentifier elements scoped to the BSN root OID and
+// rewrite only the `extension` attribute, leaving sibling attributes intact.
+// Redaction is keyed on the root OID (not on a known BSN value) so BSNs
+// echoed back by MITZ that we never sent ourselves are also scrubbed.
+var (
+	reBSNExtensionBeforeRoot = regexp.MustCompile(
+		`extension="[^"]*"(\s+(?:[A-Za-z:][A-Za-z0-9:]*="[^"]*"\s+)*root="2\.16\.840\.1\.113883\.2\.4\.6\.3")`,
```
JorisHeadease: We could also choose to return the document in the mitz component and directly alter a copy of this document, but felt a bit heavy for just redaction

component/mitz/component.go
```diff
@@ -151,10 +151,10 @@ func (c *Component) CheckConsent(ctx context.Context, authzReq xacml.AuthzReques
 		return nil, fmt.Errorf("failed to create authorization decision query: %w", err)
 	}
 
-	// Log the XML request
+	// Log the XML request with every BSN-scoped InstanceIdentifier redacted.
 	slog.DebugContext(ctx, "Sending consent check request to MITZ",
 		slog.String("endpoint", c.consentCheckEndpoint),
-		slog.String("xmlPayload", authnDecisionQueryXml))
+		slog.String("xmlPayload", xacml.RedactBSN(authnDecisionQueryXml)))
```
reinkrul, line 157: You can serialize this XML in various forms, e.g. with namespace for the extension, or with spaces around the equal sign (`extension = ""`, unprobable as it is), or using single-quotes. It works right now, but maybe we can have a test that makes sure that the output of our serialization (what is sent/logged) works with RedactBSN? That way we don't have do go deep in XML deserialization, but we can make sure it stays working even if serialization is changed.

component/pdp/policies/system/policy.rego
```diff
@@ -0,0 +1,86 @@
...
+
+# --- Helpers ---------------------------------------------------------------
+# `is_string` guards against fail-open: without it, a non-string value would
+# raise a type error from startswith, the rule body would become undefined
+# and OPA would log the decision event unmasked.
+# URN comparison is case-insensitive per RFC 8141; FHIR URIs are case-
+# sensitive per spec, but accepting case-insensitive matching here is a
+# defensive choice for a log-scrubber.
+
+is_bsn_scoped(s) if {
```
reinkrul: I wonder if we should also mask if there's no system, which is also valid;

```
?identifier=|1234567890
```

(just for good measure).

@gasperr can you think of any other edge cases?

reinkrul (changes requested):

component/pdp/policies/system/policy.rego
JorisHeadease: Resolved, leaving the comment open for @gasperr

reinkrul (approved):

Issue #477, opened 2026-04-09
PDP PIP: Document required FHIR API capabilities


reinkrul:
## Background

The PDP (Policy Decision Point) relies on a Policy Information Point (PIP) — a FHIR R4 REST API — to enrich policy decisions with data from the EHR. For the PDP to work correctly, the PIP must support specific FHIR operations and search parameters.

This issue documents the exact requirements the PIP must fulfill, so EHR vendors can validate their PIP implementation against this checklist.

### 1. Patient BSN lookup via `_elements`

**Call:**
```http
GET /Patient/{id}?_elements=identifier
```

**Requirements:**
- Must return a `Patient` resource containing the `identifier` array
- Must honor the `_elements` parameter — only `identifier` needs to be returned (minimizes patient data sent to PDP)
- The patient identifier with system `http://fhir.nl/fhir/NamingSystem/bsn` must be present and contain exactly one BSN value

**Used by:** all policies that need to resolve a FHIR patient ID to a BSN (e.g. bgz, medicatieoverdracht)

---

### 2. Implied consent lookup via `Consent?data=`

**Call:**
```http
GET /Consent?data={ResourceType}/{resourceId}
```

Example:
```http
GET /Consent?data=Task/12AF22F3-2DE5-47E1-B3CB-B053C8621F84
```

**Requirements:**
- Must support the `data` search parameter on `Consent`, matching `provision.data.reference`
- Must return all `Consent` resources whose `provision.data.reference` refers to the given resource
- Must support FHIR pagination if too many results (`Bundle.link.relation = next`); the PDP follows pagination automatically
- The returned `Consent` resources are evaluated on:
  - `status`: must be `active`
  - `scope.coding.system`: `http://nuts-foundation.github.io/nl-generic-functions-ig/CodeSystem/nl-gf-consent-scope`
  - `provision.type`: `permit` or `deny` (deny supersedes permit for equal specificity)
  - `provision.actor[].reference`: must match the requesting organization's URA
  - `provision.action[].coding`: must include code `access` from `http://terminology.hl7.org/CodeSystem/consentaction`

**Used by:** implied-consent policies (e.g. eoverdracht_sender)

---

### 3. Resource content fetch (optional, feature-flagged)

**Call:**
```http
GET /{ResourceType}/{id}
```

**Requirements:**
- Must return the full resource when `pdp.pip.resourcecontentenabled = true`
- Used for policies that need to inspect the content of a resource to make an authorization decision

[...]

PR #485, opened 2026-04-20
#477: pip documentation for a fhir server

gasperr:
- Added a capabilities summary table to the Policy Information Point section in INTEGRATION.md, listing all three FHIR endpoints the PIP server must support (patient BSN lookup, implied consent search, resource content fetch) with their required/optional status.                                                                                                                                               
  - Documented the previously undocumented Resource Content Fetch capability (GET /{ResourceType}/{id}), including when it is active (pdp.pip.resourcecontentenabled = true).                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
  - Added a FHIR R4 CapabilityStatement (docs/pip-capability-statement.json)


docs/INTEGRATION.md
```diff
@@ -589,10 +599,19 @@ To come to a policy decision the PDP might need additional information from a po
 
 Read our [configuration guide](/docs/CONFIGURATION.md) to see the options for configuring this endpoint.
 
-The PIP is a FHIR R4 REST-compatible API used for two purposes:
+The PIP must be a FHIR R4 REST-compatible API. It is called exclusively by the Knooppunt and must not be
```
reinkrul: If the EHR only has 1 FHIR API, "must not be
accessible to external parties" isn't really the right wording?

docs/INTEGRATION.md
```diff
@@ -733,6 +753,35 @@ When the PDP evaluates a request for `Task/12AF22F3-2DE5-47E1-B3CB-B053C8621F84`
...
+```json
+{
+  "resourceType": "Task",
+  "id": "12AF22F3-2DE5-47E1-B3CB-B053C8621F84",
+  "status": "requested",
+  "..."
+}
+```
+
+This capability is disabled by default and only needs to be supported when the feature flag is enabled.
```
reinkrul: And maybe more importantly, if there are authZ policies that use it (currently, none).

docs/pip-capability-statement.json
```diff
@@ -0,0 +1,52 @@
...
+  "copyright": "CC0",
+  "kind": "requirements",
+  "fhirVersion": "4.0.1",
+  "format": [
+    "json"
+  ],
+  "rest": [
+    {
+      "mode": "server",
+      "documentation": "The PIP server must be accessible only from the Knooppunt. It must support JSON format. Pagination (Bundle.link with relation 'next') must be supported for search results. When pdp.pip.resourcecontentenabled is true, the server must additionally support read on any resource type that can be targeted by incoming requests (e.g. Task, Composition). This capability is not listed per resource type below because it depends on which resource types the EHR exposes.",
```
reinkrul: >must be supported for search results

But only if the API returns paginated response? But that in itself is a bit superfluous to state, since bundle link ('next') is the way for FHIR to handle pagination?

Although, I think FHIR doesn't explicitly specify (only suggest) the bundle link relation names (`next`), so it might be worth specifying that pagination it expected to use `next` indeed?

reinkrul (changes requested):

reinkrul (approved):

PR #492, opened 2026-04-22
Fix PEP Authorization to send the request body to subrequest

melihaydogd:
The pull request does the following:
1. Removes `auth_request` subrequest to use the main request context. This is done because `auth_request` doesn't forwards request body. The request body is necessary for POST `/fhir/Patient/_search` requests since the BSN is extracted from the request body.
2. Adds `Content-Type` header in the PDP request since PDP needs it to extract POST requests body.
3. Updated the `Content-Type` check in PDP so that it can handle the values like `application/x-www-form-urlencoded; charset=UTF-8` which HAPI FHIR sends in POST search requests by default.

### DIFF
--- a/component/pdp/fhirreq.go
+++ b/component/pdp/fhirreq.go
@@ -448,7 +448,10 @@ func NewPolicyInput(request APIRequest) (*PolicyInput, error) {
 	}
 
 	var rawParams url.Values
-	hasFormData := request.Input.Request.Header.Get("Content-Type") == "application/x-www-form-urlencoded"
+	contentTypeParts := strings.Split(request.Input.Request.Header.Get("Content-Type"), ";")
+	hasFormData := slices.ContainsFunc(contentTypeParts, func(part string) bool {
+		return strings.TrimSpace(part) == "application/x-www-form-urlencoded"
+	})
 	interWithBody := []fhir.TypeRestfulInteraction{
 		fhir.TypeRestfulInteractionSearchType,
 		fhir.TypeRestfulInteractionOperation,
--- a/pep/nginx/conf.d/knooppunt.conf
+++ b/pep/nginx/conf.d/knooppunt.conf
@@ -50,15 +50,21 @@ server {
     # FHIR_BASE_PATH: incoming path clients use (default: /fhir)
     # FHIR_UPSTREAM_PATH: path on backend FHIR server (default: same as FHIR_BASE_PATH)
     location ${FHIR_BASE_PATH}/ {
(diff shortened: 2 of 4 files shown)


Dirklectisch (commented):
We should probably handle the `; charset=utf-8` case and I would like @JorisHeadease to have a look at this PR.

JorisHeadease (changes requested):
Good catch on a real bug (auth_request drops the body + charset parsing). My main concern: swapping `auth_request` for `js_content` + `internalRedirect` is an architectural change and I'd like to see it proven end-to-end before merging.

Can you extend `test/e2e/pep/authorization_test.go` with a POST `/fhir/Patient/_search` case using `Content-Type: application/x-www-form-urlencoded; charset=UTF-8` and a BSN identifier in the body? The test should assert:
1. PDP decides allow/deny based on the BSN from the body (not on an empty body falling through to default deny).
2. The FHIR backend actually receives the POST with method and body intact after `internalRedirect`.

Side notes on Copilot's comments (not blocking):
- Normalizing the Content-Type in njs at the source is cleaner than stripping `;` on both sides. Pick one place.
- The njs unit test for `authorizeAndProxy` (denied calls `request.return`, allowed calls `request.internalRedirect`) is worth adding, but it won't catch the real risk since it mocks nginx away. The e2e above is what we actually need.

JorisHeadease:
Started working on the feedback myself as it's needed/blocking poc12. I'm merging in the PR and creating a new PR for the e2e and some small improvements as I cannot puh to formelio's fork

JorisHeadease (approved):

component/pdp/fhirreq.go
```diff
@@ -448,7 +448,10 @@ func NewPolicyInput(request APIRequest) (*PolicyInput, error) {
 	}
 
 	var rawParams url.Values
-	hasFormData := request.Input.Request.Header.Get("Content-Type") == "application/x-www-form-urlencoded"
+	contentTypeParts := strings.Split(request.Input.Request.Header.Get("Content-Type"), ";")
+	hasFormData := slices.ContainsFunc(contentTypeParts, func(part string) bool {
+		return strings.TrimSpace(part) == "application/x-www-form-urlencoded"
+	})
```
reinkrul, line 454: You could use https://pkg.go.dev/mime#ParseMediaType

JorisHeadease: that was fixed @ https://github.com/nuts-foundation/nuts-knooppunt/pull/494 as I couldn't commit improvements here

PR #493, opened 2026-04-23
Use spec-compliant OAuth client assertion field names

rolandgroen:
## Summary

Renamed OAuth2 form fields `client_credentials_type`/`client_credentials` to `client_assertion_type`/`client_assertion` per RFC 7523 (JWT Bearer Token Profile). The proeftuin OAuth server may be lenient, but the spec-compliant names work against stricter implementations.

Discovered during the GF Proeftuin hackathon by comparing nuts-knooppunt against a working Python reference implementation. The other proeftuin compatibility issue (HKDF input JSON serialization) is being addressed by aligning all implementations on [RFC 8785 (JCS)](https://www.rfc-editor.org/rfc/rfc8785) — Go's default `json.Marshal` already produces JCS-compliant output for the `prsIdentifier` struct, so no knooppunt change is needed there.

## Test plan
- [x] Existing unit tests passing
- [x] OAuth assertion field rename validated against the proeftuin OAuth server (token acquisition succeeds)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/component/authn/oauth2.go
+++ b/component/authn/oauth2.go
@@ -141,8 +141,8 @@ func (t tokenSource) Token() (*oauth2.Token, error) {
 		"grant_type":              {"client_credentials"},
 		"scope":                   {strings.Join(t.scope, " ")},
 		"target_audience":         {t.targetAudience},
-		"client_assertion_type": {"urn:ietf:params:oauth:client-assertion-type:jwt-bearer"},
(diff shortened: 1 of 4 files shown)


Dirklectisch:
I think we should probably confirm with @bramwesselo what the format should be.

gasperr (approved):
Tested myself as well, so all good. Go ahead.

PR #496, opened 2026-04-24
fix(pseudonymisation): canonicalize PRS request body per RFC 8785

stevenvegt:
This PR addresses the following change in the spec: https://github.com/minvws/generiekefuncties-docs/commit/03c641ccfa33f89b8df23a4ab8b13a5a08daac2c

## Summary
- The PRS `/oprf/eval` request body was serialised with the stdlib `encoding/json`, which does not produce RFC 8785 output. This PR runs the marshalled bytes through JCS (`github.com/gowebpki/jcs`) so the body is deterministic on the wire.
- Added `github.com/gowebpki/jcs v1.0.1` (Apache-2.0, stdlib-only deps, tagged release) as a direct dependency.

## Test plan
- [x] `go build ./component/pseudonymisation/...`
- [x] `go test ./component/pseudonymisation/...` (existing `component_test.go` decodes the body by field name, unaffected by JCS's key reordering)
- [ ] Verify against a real PRS instance that it still accepts the canonicalised body

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Dirklectisch (approved):

reinkrul (changes requested):
Do we really need a library for marshalling 1 very simple struct (which we control), of 3 fields? We could literally solve this with string concatenation and save a dependency.

(as we don't use JSON canonicalization anywhere else)

stevenvegt:
> Do we really need a library for marshalling 1 very simple struct (which we control), of 3 fields? We could literally solve this with string concatenation and save a dependency.
> 
> (as we don't use JSON canonicalization anywhere else)

Yes, that is another route. I prefer the lib, but it's not a hill to die on. How bad is this "another dependency"?

reinkrul (approved):
I'll find another hill to die on.

PR #497, opened 2026-04-24
Add a `trusted` flag to mCSD administration directories

Dirklectisch:
## Summary                                                                                
                                                                                            
  Add a `trusted` flag to mCSD administration directories. A trusted directory's contents   
  are accepted without per-resource validation, intended for directories the  
  operator controls or otherwise trusts (e.g. the LRZA).                                    
                                                                                          
  Configured per directory via `mcsd.admin.<name>.trusted: true`. Defaults to `false`, so   
  existing deployments are unaffected.                                                      
                                                                                            
  ## Behavior                                                                               

  When `trusted: true` on a registered directory:

  - `ValidateUpdate` short-circuits after the `AllowedResourceTypes` check — per-resource   
  spoofing checks (URA membership, partOf chain, endpoint reference, etc.) are skipped.
  - `ValidateParentOrganizations` is skipped.                                               
  - The parent-organization map is only built when discovery is also enabled (it's the only 
  remaining consumer in trusted mode).                                                      
  - The URA-identifier-change re-query is skipped — a URA change syncs like any other change
   rather than triggering a full re-fetch.                                                  
                                                   
  Explicitly **unchanged** by `trusted`:                                                    
                                                   
  - Sync cadence: incremental sync via `_since` and `lastUpdateTimes` still applies.        
  - Discovery: a trusted root with `discover: true` still discovers and registers leaf
  directories. Discovered children are always registered as `trusted: false`.               
  - The `isDiscoverableDirectory` filter (root directories only sync mCSD-directory
  `Endpoint` resources) is unchanged.                                                       
  - `AllowedResourceTypes` is still enforced.

### DIFF
--- a/config/knooppunt.yml
+++ b/config/knooppunt.yml
@@ -9,6 +9,13 @@ mcsd:
   query:
     fhirbaseurl: "http://localhost:8080/fhir"
 
+  # Administration directories to sync from.
+  # Set `trusted: true` to skip anti-spoofing validation on a directory's contents.
+  # admin:
+  #   example:
+  #     fhirbaseurl: "https://fhir.example.org/fhir"
+  #     trusted: true
+
   # Exclude administration directories from being registered
   # Use this to prevent your own FHIR server from being added as an admin directory
   # This avoids syncing your own resources back to yourself
--- a/component/mcsd/component.go
+++ b/component/mcsd/component.go
@@ -102,6 +102,7 @@ type administrationDirectory struct {
 	fhirBaseURL      string
 	resourceTypes    []string
 	discover         bool
+	trusted          bool   // Skip validation checks on this directory's contents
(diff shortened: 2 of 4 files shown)


config/knooppunt.yml
```diff
@@ -9,6 +9,13 @@ mcsd:
   query:
     fhirbaseurl: "http://localhost:8080/fhir"
 
+  # Administration directories to sync from.
+  # Set `trusted: true` to skip anti-spoofing validation on a directory's contents.
+  # admin:
+  #   example:
+  #     fhirbaseurl: "https://fhir.example.org/fhir"
+  #     trusted: true
```
reinkrul, line 17: Or we define a directory with key/name `root` (here `example`) that is always trusted? Because in the centralized model, there will always be 1 admin directory, and it will always be trusted.

Then we can also remove the decentralized app logic in the future (when we don't need it any more), without changing the configuration structure.

PR #498, opened 2026-04-29
demo-ehr: backend proxy for FHIR APIs + dev-login bypass

reinkrul:
## Summary

- Replaces the SPA's direct FHIR fetches with a small Node/Express backend (`server.js`) that serves the production build and proxies `/api/{fhir,fhir-stu3,mcsd,knooppunt,dynamic-proxy}` to the configured upstreams. Each upstream has a per-`{method, path-regex}` allowlist of operations the frontend actually uses; everything else returns 403. The deployed demo-ehr at `/ehr/` no longer needs the browser to be able to reach the FHIR servers directly.
- Allowlist is shared between `server.js` (prod image) and `src/setupProxy.js` (CRA `npm start`) via `proxy-allowlist.js`, so dev and prod behave identically.
- Splits the Dockerfile: `Dockerfile` is now a multi-stage production image (`npm run build` → minimal Node runtime that runs `server.js`); `Dockerfile.dev` preserves the existing CRA dev-server with hot reload. `docker-compose.yml`'s `demoehr` profile builds from `Dockerfile.dev` and points at the correct `mock-components/demo-ehr/...` paths (the previous entry referenced a non-existent `./demo-ehr/...` directory).
- Refactors `api/*.js` and the two inline page fetches in `PatientPage.js` / `PatientContextLaunchPage.js` to use relative paths (`/api/fhir`, `/api/fhir-stu3`, `/api/mcsd`). The two remaining absolute-URL usages of `config.mcsdQueryBaseURL` are intentional — they embed Reference values into FHIR resources sent to other parties and must stay publicly resolvable.
- Adds a `REACT_APP_DEV_LOGIN`-gated **Dev login (skip OIDC)** button on the home page for local development. Fabricates a fixed user (`sub=00000666`, name=Dev User), persists it in `localStorage`, auto-resolves a Practitioner, and short-circuits logout. The button only renders when the env flag is set, and a stored dev session won't be restored without the flag — so it's safe to ship the code disabled.
- Drive-by fix: `Back to Dashboard` on the patients page pointed at `/mock-components/demo-ehr/public`; now `/`.

## Test plan

- [ ] `docker compose --profile demoehr up -d --build knooppunt hapi-fhir hapi-fhir-healthcheck init demo-ehr` brings up a working stack on http://localhost:3000/
- [ ] In dev mode, "Dev login (skip OIDC)" button is visible and lets you reach `/patients`
- [ ] `GET /api/mcsd/Organization` proxies to hapi-fhir and returns a Bundle
- [ ] `PATCH /api/fhir/Patient/x` and other unlisted method/path combos return 403 from the proxy (test against the production image: `docker build -f mock-components/demo-ehr/Dockerfile mock-components/demo-ehr`)
- [ ] `npm run build` in `mock-components/demo-ehr` still succeeds
- [ ] `Back to Dashboard` button on `/patients` navigates to `/`

Assisted by AI


Dirklectisch (approved):
🤞 🤞 🤞

PR #499, opened 2026-04-30
Handle delegated requests in Mitz consent check

Dirklectisch:
When a request is made under a HealthCareProfessionalDelegationCredential (mandaat), the Mitz consent query needs to identify both the delegating practitioner (legally responsible) and the delegated-to practitioner (the one actually performing the action). This change wires that through the PDP -> XACML pipeline.

### DIFF
--- a/component/mitz/xacml/types.go
+++ b/component/mitz/xacml/types.go
@@ -79,7 +79,7 @@ type AuthzRequest struct {
 	// Example: "000095254" (UZI number)
 	ProviderID string
 
-	// MandatedID UZI number of practitioner who delegated
+	// MandatedID is the UZI number of the practitioner acting under the mandate (delegated-to)
 	// Example: "000095254" (UZI number)
 	MandatedID *string
 
--- a/component/mitz/xacml/types.go
+++ b/component/mitz/xacml/types.go
@@ -79,6 +79,10 @@ type AuthzRequest struct {
 	// Example: "000095254" (UZI number)
 	ProviderID string
 
+	// MandatedID UZI number of practitioner who delegated
+	// Example: "000095254" (UZI number)
+	MandatedID *string
+
 	// ProviderInstitutionID is the unique identifier (URA code) of the healthcare institution (consulting organization)
 	// REQUIRED - Identifies the consulting/responsible organization.
 	// Example: "00000666" (institution URA code)
--- a/component/pdp/mitz.go
+++ b/component/pdp/mitz.go
@@ -60,7 +60,7 @@ func xacmlFromInput(input PolicyInput) (xacml.AuthzRequest, error) {
 		// If the request is mandated, the practitioner who mandated is responsible
 		s, _ := delRegByProp.(string)
 		iden, err := fhirutil.TokenToIdentifier(s)
-		if err != nil && iden.Value != nil {
+		if err == nil && iden.Value != nil {
 			responsiblePractitioner = *iden.Value
 		} else {
 			return xacml.AuthzRequest{}, errors.New("invalid format for delegation_registered_by")
@@ -69,7 +69,7 @@ func xacmlFromInput(input PolicyInput) (xacml.AuthzRequest, error) {
 		if hasDelegatedRole {
 			s, _ = delRoleProp.(string)
 			iden, err = fhirutil.TokenToIdentifier(s)
-			if err != nil && iden.Value != nil {
+			if err == nil && iden.Value != nil {
 				responsiblePractitionerRole = *iden.Value
 			} else {
 				return xacml.AuthzRequest{}, errors.New("missing delegation_role_code")
--- a/component/pdp/mitz.go
+++ b/component/pdp/mitz.go
@@ -2,6 +2,7 @@ package pdp
 
 import (
 	"context"
+	"errors"
 	"log/slog"
 
 	"github.com/nuts-foundation/nuts-knooppunt/component/mitz/xacml"
@@ -26,7 +27,16 @@ func (c *Component) enrichPolicyInputWithMitz(ctx context.Context, input *Policy
 		return input, resultReasons
 	}
 
-	consentReq := xacmlFromInput(*input)
+	consentReq, err := xacmlFromInput(*input)
+	if err != nil {
+		slog.WarnContext(ctx, "Mitz consent check failed", "error", err)
+		return input, []ResultReason{
+			{
+				Code:        TypeResultCodeUnexpectedInput,
+				Description: "could not complete consent check with Mitz: " + err.Error(),
+			},
+		}
+	}
 	consentResp, err := c.consentChecker.CheckConsent(ctx, consentReq)
 	if err != nil {
 		slog.WarnContext(ctx, "Mitz consent check failed", "error", err)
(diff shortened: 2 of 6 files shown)


component/pdp/mitz_test.go
```diff
@@ -24,14 +25,101 @@ func TestComponent_map_input_xacml(t *testing.T) {
...
+	assert.Equal(t, "TREAT", req.PurposeOfUse)
+	assert.Nil(t, req.MandatedID, "MandatedID should not be set when no delegation is present")
+}
+
+func TestComponent_xacmlFromInput_Delegation(t *testing.T) {
+	const (
+		delegatorID   = "000012345"
+		delegatorRole = "30.000"
+		regByToken    = "http://fhir.nl/fhir/NamingSystem/uzi-nr-pers|" + delegatorID
+		roleCodeToken = "http://fhir.nl/fhir/NamingSystem/role-code-pers|" + delegatorRole
```
Dirklectisch: Unsure what the actual systems here will be, is this delegation credential documented somewhere? I couldn't find it in [our credential catalog](https://nuts-foundation.github.io/nl-generic-functions-ig/credential-catalog.html).

Dirklectisch: Answering my own question, the system for role code is: `http://fhir.nl/fhir/NamingSystem/uzi-rolcode`

component/pdp/mitz_test.go
```diff
@@ -24,14 +25,101 @@ func TestComponent_map_input_xacml(t *testing.T) {
...
+	assert.Equal(t, "TREAT", req.PurposeOfUse)
+	assert.Nil(t, req.MandatedID, "MandatedID should not be set when no delegation is present")
+}
+
+func TestComponent_xacmlFromInput_Delegation(t *testing.T) {
+	const (
+		delegatorID   = "000012345"
+		delegatorRole = "30.000"
+		regByToken    = "http://fhir.nl/fhir/NamingSystem/uzi-nr-pers|" + delegatorID
+		roleCodeToken = "http://fhir.nl/fhir/NamingSystem/uzi-rolcode" + delegatorRole
```
reinkrul: **Missing `|` separator — this breaks CI.**

`TokenToIdentifier` requires a `system|value` shape and rejects this token as malformed, so the first sub-test (`uses delegating practitioner as responsible…`) fails. Should be:

```go
roleCodeToken = "http://fhir.nl/fhir/NamingSystem/uzi-rolcode|" + delegatorRole
```

_Assisted by AI_

component/mitz/xacml/types.go
```diff
@@ -79,6 +79,10 @@ type AuthzRequest struct {
 	// Example: "000095254" (UZI number)
 	ProviderID string
 
+	// MandatedID UZI number of practitioner who delegated
```
reinkrul: **Doc comment is inverted.**

The call site sets `MandatedID = input.Subject.User.Id` — the authenticated user, who is the practitioner the mandate was delegated _to_, not the one who delegated. The URN `urn:nl:otv:names:tc:1.0:subject:mandated` likewise refers to the mandate holder. Suggest:

```go
// MandatedID is the UZI number of the practitioner acting under the mandate (delegated-to).
```

_Assisted by AI_

component/pdp/shared.go
```diff
@@ -142,6 +142,10 @@ type PolicySubjectUser struct {
 	Id   string `json:"id"`
 	Role string `json:"role"`
 }
+type PolicySubjectDelegation struct {
```
reinkrul: **Dead code.**

`PolicySubjectDelegation` is added but never referenced anywhere in the diff. Either wire `PolicySubject` to use it (instead of fishing strings out of `OtherProps["delegation_registered_by"]` / `"delegation_role_code"` in `mitz.go`), or drop the type. As-is it's just noise.

_Assisted by AI_

component/pdp/mitz.go
```diff
@@ -39,19 +51,59 @@ func (c *Component) enrichPolicyInputWithMitz(ctx context.Context, input *Policy
...
+			return xacml.AuthzRequest{}, errors.New("invalid format for delegation_registered_by")
+		}
+
+		if hasDelegatedRole {
+			s, _ = delRoleProp.(string)
+			iden, err = fhirutil.TokenToIdentifier(s)
+			if err == nil && iden.Value != nil {
+				responsiblePractitionerRole = *iden.Value
+			} else {
+				return xacml.AuthzRequest{}, errors.New("missing delegation_role_code")
```
reinkrul: **Misleading error: "missing" vs "malformed" collapsed.**

Both the absent (line 80) and malformed (this line) cases return `"missing delegation_role_code"`. Operationally these are different problems — mirror how `delegation_registered_by` is handled and return `"invalid format for delegation_role_code"` here.

_Assisted by AI_

component/pdp/mitz.go
```diff
@@ -39,19 +51,59 @@ func (c *Component) enrichPolicyInputWithMitz(ctx context.Context, input *Policy
 	return input, nil
 }
 
-func xacmlFromInput(input PolicyInput) xacml.AuthzRequest {
-	return xacml.AuthzRequest{
+func xacmlFromInput(input PolicyInput) (xacml.AuthzRequest, error) {
+	delRegByProp, isDelegated := input.Subject.OtherProps["delegation_registered_by"]
```
reinkrul, line 54: **Asymmetric: `delegation_role_code` without `delegation_registered_by` is silently ignored.**

The whole delegation branch is gated on `isDelegated` (presence of `delegation_registered_by`). If a token has only `delegation_role_code`, we silently treat the request as non-delegated and use the authenticated user as `ProviderID`. That introspection state is internally inconsistent and should probably be rejected explicitly.

_Assisted by AI_

component/pdp/mitz.go
```diff
@@ -73,19 +125,19 @@ func validateMitzInput(input PolicyInput) []ResultReason {
 		},
 		{
 			Value:   input.Subject.User.Role,
```
reinkrul, line 130: **`User.Role` is still required here, but the delegation path overwrites it.**

In the delegated path, `SubjectRole` comes from `delegation_role_code`, not `User.Role`. This validator still rejects the request if `User.Role` is empty. Tolerable today (introspection always populates it), but worth either a comment or expanding `validateMitzInput` to cover the delegation-specific fields explicitly so future changes don't fall through the cracks.

_Assisted by AI_

component/pdp/mitz.go
```diff
@@ -39,19 +51,59 @@ func (c *Component) enrichPolicyInputWithMitz(ctx context.Context, input *Policy
...
+		SubjectRole:            responsiblePractitionerRole,
+		ProviderID:             responsiblePractitioner,
 		ProviderInstitutionID:  input.Subject.Organization.Ura,
 		ConsultingFacilityType: input.Subject.Organization.FacilityType,
 		PurposeOfUse:           "TREAT",
 	}
+
+	if isDelegated {
+		// If the request is mandated add the practitioner that has been delegated to
+		req.MandatedID = to.Ptr(input.Subject.User.Id)
```
reinkrul: **Use `new(string)` instead of `to.Ptr`.**

Pulling in `caramel/to` for one pointer-wrap is heavy when the stdlib already covers it:

```go
req.MandatedID = new(input.Subject.User.Id)
```

_Assisted by AI_

component/pdp/mitz_test.go
Dirklectisch: Typo fixed

component/mitz/xacml/types.go
Dirklectisch: Makes sense

component/pdp/shared.go
Dirklectisch: Old code indeed

component/pdp/mitz.go
Dirklectisch: 👍

component/pdp/mitz.go
Dirklectisch: Added an extra clause

component/pdp/mitz.go
Dirklectisch: Let's keep it like its is right now, I don't want to reimplement all the edge cases in the validation. It just detects early all required fields for better error handling.

component/pdp/mitz.go
Dirklectisch: My editor doesn't understand this syntax yet even when I point it to the latest GO SDK and reindex. In any case it does compile so let's roll with this suggestion.

reinkrul (approved):

gasperr:
Tested locally with a configured Mitz integration.

Tested with

```
{
  "input": {
    "subject": {
      "user_id": "000095254",
      "user_role": "01.015",
      "organization_ura": "01234567",
      "organization_facility_type": "Z3",
      "scope": "bgz"
    },
    "request": {
      "method": "GET",
      "protocol": "HTTP/1.0",
      "path": "/Coverage",
              "query": "_include=Coverage%3Apayor%3APatient&_include=Coverage%3Apayor%3AOrganization"
    },
    "context": {
      "data_holder_organization_id": "01234567",
      "data_holder_facility_type": "Z3",
      "connection_type_code": "hl7-fhir-rest",
      "patient_bsn": "900203638"
    }
  }
}
```

got back 
```
{
    "allow": true,
    "policies": {
        "bgz": {
            "allow": true,
            "reasons": null
        }
    }
}
```

with mitz check triggering and returning a Permit response to the closed question query.

gasperr (approved):

Dirklectisch:
Thanks for testing @gasperr!

Issue #370, opened 2026-02-05
PDP: Support FHIR STU3


reinkrul:
We now converted all FHIR CapabilityStatements to FHIR R4, but at some point it'll break, leading to a PDP that can't authorize a STU3 request that has no equivalent in R4.

We probably do not need to update the Go lang FHIR model library to support STU3 but rather choose to parse the capability statements in another way. Either: use a json parser instead of the model library OR use a rego policy to express the capability statement check.

This is not high priority for the PoCs but would be good to do anyway if we run out of work.

PR #507, opened 2026-05-23
PDP STU3 capability statement parsing

gasperr:
#370 
PDP can now parse stu3 capability statements as well as r4 ones


reinkrul (changes requested):
Can we also revert back to the official CapabilityStatement for BgZ? I think we took the official one and translated it to R4

reinkrul (approved):

PR #521, opened 2026-06-10
New addressing component that only syncs the centralised LRZA per the current IG from VWS

Dirklectisch:
This PR includes and depends on the work that is also available for review here:

https://github.com/nuts-foundation/nuts-knooppunt/pull/520

Adds a new `lrza` component that synchronizes the trusted national **LRZA** (Landelijk Register Zorgaanbieders) mCSD directory into a local mCSD query directory.

The existing `mcsd` component is built for peer-to-peer discovery with anti-spoofing validation. The LRZA is a single, government-controlled, fully-trusted source, so that machinery is unnecessary. Rather than fold a "trusted" mode into `mcsd`, this adds a dedicated, simpler component and leaves `mcsd`'s flow essentially untouched.

The PR contains tests but I've also verified myself that it works against the current LRZA implementation in the iRealisatie proeftuin.

## Known limitations / follow-ups

- **Large directories:** the sync currently applies as a single FHIR transaction; the per-sync entry cap was set to 10000 as an interim measure. Splitting into bounded chunks is non-trivial because FHIR servers that enforce referential integrity on write require referenced resources to exist first — so safe chunking needs dependency-ordered batching. If we want that I say we take it as a separate task.


cmd/start.go
```diff
@@ -52,6 +53,17 @@ func Start(ctx context.Context, config Config) error {
...
 
+	// The LRZA sync client is only registered when a trusted source directory is configured.
+	if config.LRZA.LRZABaseUrl != "" {
+		lrzaClient, err := lrza.New(config.LRZA)
+		if err != nil {
+			return errors.Wrap(err, "failed to create LRZA sync client")
+		}
+		components = append(components, lrzaClient)
+	} else {
+		slog.InfoContext(ctx, "LRZA sync client is disabled (no source directory configured)")
```
reinkrul: "source directory" sounds like something on disk, as it's centralized, just naming for what it is might be better; "LRZA base URL not configured"

component/lrza/component.go
```diff
@@ -0,0 +1,518 @@
...
+	CountDeleted int      `json:"deleted"`
+	Warnings     []string `json:"warnings"`
+	Errors       []string `json:"errors"`
+}
+
+// Component syncs a single trusted mCSD Directory into the local query directory.
+type Component struct {
+	config          Config
+	sourceClient    fhirclient.Client
+	fhirQueryClient fhirclient.Client
```
reinkrul: I think 2 classes of terminology are mixed here; I'd expect `targetClient` and `sourceClient` or `fhirQueryClient` and `fhirAdministrationClient`. Probably the first pair, for simplicity

component/lrza/component.go
```diff
@@ -0,0 +1,518 @@
...
+			return nil, fmt.Errorf("LRZA mTLS is configured but failed to load: %w", err)
+		}
+		baseTransport = &http.Transport{TLSClientConfig: tlsConfig}
+		slog.Info("LRZA: mTLS configured for source connection", slog.String("certFile", config.TLSCertFile))
+	}
+
+	tracedTransport := tracing.WrapTransport(baseTransport)
+
+	if config.Auth.IsConfigured() {
+		slog.Info("LRZA: OAuth2 authentication configured", slog.String("token_endpoint", config.Auth.TokenEndpoint))
```
reinkrul: not really relevant INFO logging, only clutter PRD logging. Maybe log the other way around (as it's clearly possible to not use auth for testing/development purposes?)

lib/fhirutil/history.go
```diff
@@ -0,0 +1,89 @@
...
+
+// DeduplicateHistoryEntries collapses a FHIR _history feed so each resource appears at most once,
+// keeping only its most recent version. A _history query can return several versions of the same
+// resource (including a DELETE), but a transaction Bundle must reference each resource at most once,
+// so the feed must be deduplicated before it can be replayed as a transaction.
+//
+// FHIR _history entries are full snapshots rather than deltas, so keeping the latest version yields
+// the resource's current state; versions are not merged.
+//
+// The implementation relies on the spec-mandated history ordering: a history Bundle is "sorted with
```
reinkrul, line 15: I think this is dangerous to assume, maybe this function should do active ordering (or keep track of latest with date)?

lib/fhirutil/history.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"github.com/zorgbijjou/golang-fhir-models/fhir-models/fhir"
+)
+
+// DeduplicateHistoryEntries collapses a FHIR _history feed so each resource appears at most once,
+// keeping only its most recent version. A _history query can return several versions of the same
+// resource (including a DELETE), but a transaction Bundle must reference each resource at most once,
+// so the feed must be deduplicated before it can be replayed as a transaction.
+//
+// FHIR _history entries are full snapshots rather than deltas, so keeping the latest version yields
```
reinkrul, line 12: unless a JSON Patch operation was used, maybe explicitly check for that? (and error out)

Are there more edge cases?

lib/fhirutil/history.go
```diff
@@ -0,0 +1,89 @@
...
+// drop the rest - no timestamp comparison needed - which also makes DELETEs (the newest version,
+// hence first) correctly win. Callers querying _history must not override that ordering. Input order
+// is preserved in the output.
+//
+// Resources are keyed by "ResourceType/id", not by bare id: a FHIR logical id is only unique within
+// a resource type, so e.g. Organization/1 and Endpoint/1 are distinct resources that must not be
+// collapsed into one.
+func DeduplicateHistoryEntries(entries []fhir.BundleEntry) []fhir.BundleEntry {
+	seen := make(map[string]struct{}, len(entries))
+	result := make([]fhir.BundleEntry, 0, len(entries))
```
reinkrul, line 27: not sure if this is the right initialization size; if you have a very large history of just a couple of resources (that changed often), this allocates far too much. What is the expected use case here? Many updates, or little updates?

component/lrza/component.go
```diff
@@ -0,0 +1,518 @@
...
+// The same base transport - including mTLS - is reused for OAuth2 token requests.
+func newSourceHTTPClient(config Config) (*http.Client, error) {
+	var baseTransport http.RoundTripper = http.DefaultTransport
+	if config.TLSCertFile != "" {
+		tlsConfig, err := tlsutil.CreateTLSConfig(config.Config)
+		if err != nil {
+			return nil, fmt.Errorf("LRZA mTLS is configured but failed to load: %w", err)
+		}
+		baseTransport = &http.Transport{TLSClientConfig: tlsConfig}
+		slog.Info("LRZA: mTLS configured for source connection", slog.String("certFile", config.TLSCertFile))
```
reinkrul: I'd expect this to be the default flow, so no. need to be explicit about it (log exceptions, not normal behavior)

component/lrza/component.go
```diff
@@ -0,0 +1,518 @@
...
+}
+
+// update runs one sync cycle: fetch the trusted source's history, build a transaction from it, apply
+// it to the query directory, and record the timestamp for the next incremental sync.
+func (c *Component) update(ctx context.Context) (UpdateReport, error) {
+	c.updateMux.Lock()
+	defer c.updateMux.Unlock()
+
+	run := c.newSyncRun()
+	slog.InfoContext(ctx, "Updating from trusted LRZA directory",
```
reinkrul: "trusted" -> central?

might be clearer for operators/implementors

component/lrza/component.go
```diff
@@ -0,0 +1,518 @@
...
+		logging.FHIRServer(c.config.LRZABaseUrl),
+		slog.Any("resourceTypes", c.resourceTypes),
```
reinkrul: This doesn't change (resourceTypes) so should we be logging it? It'll only change if someone reconfigures it (because...?) As it's part of structured log, it might also be indexed, but it's not really a useful field to search on imo. (Same could be argued for LRZA base URL).

We could also log it once, at component start

component/lrza/component.go
```diff
@@ -0,0 +1,518 @@
...
+	var searchSet fhir.Bundle
+	if err := c.sourceClient.SearchWithContext(ctx, "", searchParams, &searchSet, fhirclient.AtPath(path)); err != nil {
+		return nil, fhir.Bundle{}, fmt.Errorf("search of %s failed: %w", path, err)
+	}
+
+	var entries []fhir.BundleEntry
+	err := fhirclient.Paginate(ctx, c.sourceClient, searchSet, func(set *fhir.Bundle) (bool, error) {
+		entries = append(entries, set.Entry...)
+		if len(entries) >= maxUpdateEntries {
+			return false, fmt.Errorf("too many entries (%d), aborting update to prevent excessive memory usage", len(entries))
```
reinkrul, line 279: is there a fix for this? Say, if the pilot runs for a few months, and the number of entries exceeds what's allowed, what would be the fix the operator can do to resolve it?

component/lrza/component.go
```diff
@@ -0,0 +1,518 @@
...
+	run.entries = libfhir.DeduplicateHistoryEntries(entries)
+	return nil
+}
+
+// queryResourceType queries a single resource type, following pagination up to maxUpdateEntries. It
+// reads the _history endpoint for an incremental sync, or searches the resource type directly for the
+// initial full sync.
+func (c *Component) queryResourceType(ctx context.Context, run *syncRun, resourceType string, searchParams url.Values) ([]fhir.BundleEntry, fhir.Bundle, error) {
+	path := resourceType
+	if run.incremental {
```
reinkrul: Trivial: passing the entire `syncRun` struct as pointer, only to use `.incremental`, seems a bit excessive (could pass `incremental bool` instead). General approach i.m.o. should be pass the simplest objects/values a function needs, rather than complex objects (for most part unused). Although: it's an internal function, so tight coupling isn't that bad.

component/lrza/component.go
```diff
@@ -0,0 +1,518 @@
...
+		"_sort": []string{"-_lastUpdated"},
+	}
+	incremental := c.lastUpdateTime != ""
+	if incremental {
+		params.Set("_since", c.lastUpdateTime)
+	}
+	return &syncRun{
+		queryStart:   time.Now(),
+		searchParams: params,
+		incremental:  incremental,
```
reinkrul: you could even make `incremental` a function that checks for the existence of the `_since` parameter, since now the state is spread over 2 values (param and incremental var)

component/lrza/component.go
```diff
@@ -0,0 +1,518 @@
...
+		return fmt.Errorf("failed to unmarshal resource: %w", err)
+	}
+	resourceType, ok := resource["resourceType"].(string)
+	if !ok {
+		return errors.New("resource has no resourceType")
+	}
+	resourceID, ok := resource["id"].(string)
+	if !ok {
+		return errors.New("resource has no id")
+	}
```
reinkrul: This feels somewhat duplicate with `ExtractResourceInfo()`, but we're also using the map here to alter the resource itself. Maybe we should make `ExtractResourceInfo()` return the unmarshalled map? Then we can re-use it here as well.

component/lrza/component.go
```diff
@@ -0,0 +1,518 @@
...
+	var entries []fhir.BundleEntry
+	for i, resourceType := range c.resourceTypes {
+		curr, searchSet, err := c.queryResourceType(ctx, run, resourceType, cloneValues(run.searchParams))
+		if err != nil {
+			return fmt.Errorf("failed to query %s: %w", resourceType, err)
+		}
+		entries = append(entries, curr...)
+		if i == 0 {
+			run.firstSearchSet = searchSet
+		}
```
reinkrul: complicated code design going on here, we just want the first LastUpdated meta timestamp. Can't introduce a var `run.queryStartedAt` and directly set it here from the meta? Storing a searchSet just to grab this field from it later on, feels like a detour

component/lrza/component.go
```diff
@@ -0,0 +1,518 @@
...
+	}
+	return nil
+}
+
+// cloneValues returns a shallow copy of the given url.Values, so per-request mutations don't affect
+// the shared run search parameters.
+func cloneValues(values url.Values) url.Values {
+	out := make(url.Values, len(values))
+	for k, v := range values {
+		out[k] = v
```
reinkrul: it still copies the slice pointer, could use `out[k] = append([]string(nil), v...)` (or something) for a deep copy, if it's a potential problem.

reinkrul (changes requested):

cmd/start.go
Dirklectisch: Fixed the text to reflect your suggestion.

component/lrza/component.go
Dirklectisch: I went with `fhirLRZAClient` and `fhirQueryClient`.

component/lrza/component.go
Dirklectisch: Currently the OAuth server in the iRealisatie proeftuin doesn't even delegate access to the centralised LRZA there. In the real situation in most likely will however. What I'll do is remove this log message and add a comment about this in the code.

lib/fhirutil/history.go
Dirklectisch: The current implementation (on main) of this function orders manually by `lastUpdated`. Deletes however have an empty body and thus can not be ordered in this way, leading to subtle bugs.

I think it's fine to assume this ordering as the FHIR spec mandates it. There is also the sort parameter being passed to the FHIR endpoint that makes explicit what the ordering should be (even though the default is already correct).

```go
params := url.Values{
	"_count": []string{strconv.Itoa(searchPageSize)},
	// Pin newest-first ordering: deduplication relies on it (a history Bundle is sorted with
	// oldest versions last). Don't trust the server default.
	"_sort": []string{"-_lastUpdated"},
}
```

lib/fhirutil/history.go
Dirklectisch: I think we can rely on the whole content of the resource being there as that what the spec says.

> Each entry SHALL minimally contain at least one of: a resource which holds the resource as it is at the conclusion of the interaction, or a request with entry.request.method The request provides information about the result of the interaction that led to this new version, and allows, for instance, a subscriber system to differentiate between newly created resources and updates to existing resources.

See: https://hl7.org/fhir/http.html#history

In any case I wouldn't know how to explicitly check if what we got is the full resource other than querying each resource in the history feed for its full state. So not sure I understand how your suggestion can be implemented.

lib/fhirutil/history.go
Dirklectisch: Why does this allocate way too much? The capacity of `seen` and `result` scale with the amount of entries in the history query. To me it seems likely that most times there will be no duplicates in the history query and thus we need exactly that capacity. When duplicates are found we expect the capacity to be less but I think that's rare.

I think using make with zero length and a specified capacity is the correct way to allocate here as we don't know the exact size, nor the exact values and we are sure this never stays empty.

component/lrza/component.go
Dirklectisch: Agree, I removed the message.

component/lrza/component.go
Dirklectisch: Agree, changed the text.

component/lrza/component.go
Dirklectisch: I think there is some use to having some kind of log message that the sync has kicked off. Use case would be during debugging or having some telemetry if the update has been triggered. I've moved logging the `LRZABaseURL` & `resourceTypes` to when the component starts.

component/lrza/component.go
Dirklectisch: There is no fix for this at the moment but we should look into implementing the use case where the LRZA has more than 10.000 items. I made note of this limitation in the PR description.

> Large directories: the sync currently applies as a single FHIR transaction; the per-sync entry cap was set to 10000 as an interim measure. Splitting into bounded chunks is non-trivial because FHIR servers that enforce referential integrity on write require referenced resources to exist first — so safe chunking needs dependency-ordered batching. If we want that I say we take it as a separate task.

As it says there safe chunking of the transaction bundle requires building up a dependency tree. If we think want to go ahead and implement that I suggest we do so in a separate PR.

component/lrza/component.go
Dirklectisch: In the other MCSD module we had a lot of long function calls with all sorts of parameters, I didn't like that. Personally I found this approach way more readable, bundling the state for a single run into a struct and passing that to each step. I can change it if you want but then we should unpack the arguments for each step in the process.

component/lrza/component.go
Dirklectisch: Changed to reflect this suggestion

component/lrza/component.go
Dirklectisch: Agree. Implemented the change, we still needed those checks though so the code looks almost the same.

component/lrza/component.go
Dirklectisch: Now directly recorded on the run as `sourceLastUpdated` instead of storing the whole result set.

component/lrza/component.go
Dirklectisch: I'm surprised you caught this! Made the change.

reinkrul: and even without AI!

reinkrul: (copying slices in Go is a real footgun, and also the notion of "shallow copy" grabbed my attention)

component/lrza/component.go
reinkrul: It's fine, it's internal anyways.

component/lrza/component.go
reinkrul: I agree, let's make sure we have an issue on the backlog for it?

lib/fhirutil/history.go
reinkrul: ok

lib/fhirutil/history.go
reinkrul: It's only used in controlled use cases, and it's a super-edge case. Not too important for now.

component/lrza/component.go
```diff
@@ -0,0 +1,527 @@
...
+func (c *Component) fetchEntries(ctx context.Context, run *syncRun) error {
+	var entries []fhir.BundleEntry
+	for i, resourceType := range c.resourceTypes {
+		curr, searchSet, err := c.queryResourceType(ctx, run, resourceType, cloneValues(run.searchParams))
+		if err != nil {
+			return fmt.Errorf("failed to query %s: %w", resourceType, err)
+		}
+		entries = append(entries, curr...)
+		if i == 0 && searchSet.Meta != nil {
+			run.sourceLastUpdated = searchSet.Meta.LastUpdated
```
reinkrul, line 254: what about empty string? (is non-nil)

reinkrul (approved):
one comment left

component/lrza/component.go
Dirklectisch: Create here:

https://github.com/nuts-foundation/nuts-knooppunt/issues/522

component/lrza/component.go
Dirklectisch: Introduced the empty string check.

